PR #16, opened 2014-05-09
Ability to add destination path files options field.

paolooo:
I've set an ability to add object { src: '...', dest: '...' } files option field. See example below.

```bash
var nw = new NwBuilder({
        files: [
           {
              src: path.resolve(__dirname, '../path/to/another/lib/**/*'),  # /home/user/me/path/to/another/lib/**/*
              dest: 'lib'  # app.nw/lib
            }
        ],
        ...
});
```

During compression or zipping, all the files from `../path/to/another/lib/**/*` will be stored to `app.nw/lib` location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,13 @@ During development you can run node-webkit with `nwbuild -r path/to/your/younwap
 Or use the module:
 
 ```js
-var NwBuilder = require('node-webkit-builder');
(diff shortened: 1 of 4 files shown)


adam-lynch:
I feel like this shouldn't be handled by this module.

PR #73, opened 2014-09-08
CLI options that correspond directly to API options

### LINKED ISSUE #67: Match all options in API, CLI and `package.json`
So that we can just write up an npm script for building instead of using Grunt/Gulp or the programmatic API.

1j01:
Fixes #67 Add additional CLI options that correspond directly to API options, such as `--appName`

This also deprecates `buildType` in favor of `buildNamingScheme`, which is, I think, a much better name. It should be backwards-compatible.

I have tested this with a few options a bit on windows, but some of these options are mac-only or not very interesting to test (cacheDir). It seems to work pretty well.

What do you think about the descriptions? I mostly took the first sentences from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Usage: nwbuild [options] [path]
 Options:
   -p, --platforms      Platforms to build, comma-sperated, can be: win,osx,linux32,linux64   [default: "osx,win"]
   -v, --version        The nw version, eg. 0.8.4                                             [default: "latest"]
-  -r, --run            Runs node-webkit for the current platform                            [default: false]  
+  -r, --run            Runs node-webkit for the current platform                             [default: false]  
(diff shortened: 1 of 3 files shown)


adam-lynch:
@steffenmllr should this be closed?

1j01:
@adam-lynch I suppose so.

julmot:
I need `buildType` as a CLI option.

1j01:
I'm still hesitant about duplicating the potential usage scope of an option that ought to be renamed, and I'm still no longer using this module. I never even used the CLI, I just thought I would try to pick some low-hanging fruit and improve parity between the CLI and API. I still feel like `buildType` "ought to" be deprecated first, but the functionality probably matters more. Hopefully someone else will pick this up.

PR #89, opened 2014-10-14
properly handle array globs with .run()

kolodny:


lib/index.js
```diff
@@ -430,7 +430,11 @@ NwBuilder.prototype.runApp = function () {
 
         self.emit('log', 'Launching App');
         return new Promise(function(resolve, reject) {
-            var p = spawn(executable, ['--enable-logging', self.options.files.replace(/\*[\/\*]*/,"")].concat(self.options.argv));
+            var baseDir = self.options.files;
+            if (baseDir instanceof Array) {
+                baseDir = baseDir[0];
+            }
+            var p = spawn(executable, ['--enable-logging', baseDir.replace(/\*[\/\*]*/,"")].concat(self.options.argv));
```
adam-lynch: The existing assumption seems bad, right? This kind of exaggerates the issue. Like if I pass `['!abc/*', 'def/*']` then the wrong baseDir will be used?

PR #93, opened 2014-10-23
Copy nw locales

### LINKED ISSUE #88: Locales are not copied over
The `locales` directory of node-webkit is not copied over with the built version, at least not on windows. Without it, `navigator.language` is empty and i10n becomes a problem :)

y-lohse:
Hi,
This change would fix #88  Without those files, `navigator.language` is empty, at least on windows.


senshu:
This fix only works for Windows. The same modification should be applied to the Linux build.

Issue #103, opened 2014-11-03
No such file or directory


kaytrance:
Have installed node-webkit-builder by _npm install node-webkit-builder -g_.

When just typing **nwbuild** in terminal I get 

```
: No such file or directory
```

I get the same output when running my script to bundle my app, here's the code:

```
#!/bin/bash
nwbuild --platforms win --buildDir "release/win" --forceDownload ~/workspace/app
```

PR #105, opened 2014-11-09
Fix #75. Show resolved sources files to avoid issues like #103. Unify coding style

### LINKED ISSUE #75: env: node\r: No such file or directory
Can't run nwbuild

It returns env: node\r: No such file or directory with every execution

nwbuild -r .
env: node\r: No such file or directory

nwbuild -h
env: node\r: No such file or directory

richardgrey:
Issues #75 fixed by replacing bin file with new one. To avoid problems with line endings in future, `end_of_line` directive was added to .editorconfig.


vikbez:
:+1: :kiss:

vikbez:
:+1: please fix this :heart:

Issue #142, opened 2014-12-18
1.0.2 seems to be broken somewhat


adam-lynch:
It seems #140 did break something after all... my head is pounding since I was up til 4am working on something. Could you see what's up with it @bastimeyer?

When building on Mac, I get an `ENOENT` error when its copying the `.icns` icon.

When building on Windows, `fs.copy` in `utils.copyFile` is giving an `EMFILE` error.

In both cases above I'm just trying to build for `osx32` only. It builds fine for Windows I think, at least when building on Windows.

Going back to 1.0.0 when building on Mac means it builds fine but there's no icon, I think. And going back to 1.0.0 has no effect on Windows, I think.

The windows problem may be a separate issue altogether but both are problems with copying. It's very strange.

cc @lucienimmink, @cbabos

adam-lynch:
@bastimeyer sorry, was very busy and sick. I assume your npm cache didn't hold up that long and running `npm install grunt-node-webkit-builder` now would give you the latest node-webkit-builder as a dependency?

adam-lynch:
1.0.8 published. It helps but I don't think it's 100% fixed. I still get EMFILE errors personally.

adam-lynch:
Closing. I'll reopen if this happens for anyone still.

PR #148, opened 2015-01-07
Fix copyFile method

### LINKED ISSUE #147: Utils.js: function copyFile always prints "chmod <perms> on <file> failed after copying, ignoring".
solution is to check if there actually was an error, ex. 

```Javascript
if (err) {
    _event.emit('log', 'chmod ' + stats.mode + ' on ' + dest + ' failed after copying, ignoring');
}
```

bambooCZ:
When file doesn't exist after copy, DO NOT COPY IT AGAIN. We will only check its existence again.

fixes #147

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -83,33 +83,34 @@ module.exports = {
     },
     copyFile: function (src, dest, _event) {
         var copy = function(resolve, reject, retryCount){
-            var onFailure = function(err){
-                if(retryCount > 2){
-                    return reject(err);
-                }
-                else {
-                    setTimeout(function(){
-                        copy(resolve, reject, retryCount + 1);
-                    }, 200);
-                }
-            };
 
             var stats = fs.lstatSync(src);
             fs.copy(src, dest, function (err) {
                 if(err) return reject(err);
 
-                fs.exists(dest, function(exists){
+                var retryCount = 0;
+                var existsCallback = function(exists){
                     if(exists){
                         fs.chmod(dest, stats.mode, function(err){
                             // ignore error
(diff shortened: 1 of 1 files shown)


adam-lynch:
We've been having some serious problems around this. See #142. 

When you say "Fix copyFile method", what exactly does this fix? Because I think there might be two / three problems overall.

bambooCZ:
Acording #142, I had the same problem with .icns file. This is fixed in this fix.

IMO, the problem is that when a file is not in destination after copy, another copying is executed.
Solution is not to execute copying again, but only check its existence later.

It works for me. I am using my modification and I haven't encountered this problem again.

bambooCZ:
But yes this is just a workaround. This should be fixed in graceful-fs-extra.copy. It should not call the callback before the file is truly there.

swissmanu:
this fix work for me on osx 10.10 as also on my travis ci build... would really appriciate a merge to the official version of `node-webkit-builder`

dubcanada:
This fixed my issue on Windows.

steida:
Any progress?

bambooCZ:
Are there any problems with it? Adam, have you tested it? Do you need any explanation? Or what is taking you so long?

adam-lynch:
Completely forgot to test it. I'm crazy busy. I'll try test this tonight I swear :smile:.

steida:
Awesome.

adam-lynch:
I still get the EMFILE myself but this does look like an improvement so I'm merging it in.

steida:
emfile? ulimit -n 10000 does not work?

adam-lynch:
@steida there's no `ulimit` command on Windows.

jminnick:
It still seems like there is some type of threshold limit on windows. I can reproduce this "Error: EMFILE, open" issue reliably.

Issue #141, opened 2014-12-18
NSHumanReadableCopyright shouldn't be mandatory


adam-lynch:
I assume at least

trevorah:
Fixed by https://github.com/mllrsohn/node-webkit-builder/commit/77c4383249d9fce17fd9dd18d44c3cd5a2f4c6e9.

PR #155, opened 2015-01-14
fixed broken osx about screen when copyright not supplied

trevorah:
Fun fact: trying to google for issues with "osx about screen" will just give you info *about* osx screens. :disappointed:

Anyway...

If you dont set a copyright, then the "My App" -> "About My App" window in osx will not open on the first click, and on the second click will give you this:

broken
Note the weird version string even if the version number is set.

If you supply the copyright info, then everything works and you get this:
copyright
Which is all hunky dory.

The problem is caused by `NSHumanReadableCopyright` being set to `false` in the Info.plist if the copyright info isnt supplied to node-webkit-builder. This causes osx to cry as it expects a string. From the [spec](https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html):

> NSHumanReadableCopyright (String - OS X). This key contains a string with the copyright notice for the bundle; for example, © 2008, My Company. You can load this string and display it in an About dialog box. This key can be localized by including it in your InfoPlist.strings files. This key replaces the obsolete CFBundleGetInfoString key.

Also note that `NSHumanReadableCopyright` hasnt been marked as "required" or even "recommended", so we dont even need to set that key if nothing is supplied to node-webkit-builder. So this fix just removes the fallback that sets the key to `false`. With this fix, this is what the about screen now looks like if you dont supply copyright info:

nocopyright
:tada:

/cc @bastimeyer as this fix undoes changes made in #143

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -190,8 +190,9 @@ module.exports = {
             obj.CFBundleVersion = parsedParams.version;
             obj.CFBundleShortVersionString = 'Version ' + parsedParams.version;
         }
-        // Read a user defined copyright text or set it to false otherwise
-        obj.NSHumanReadableCopyright = parsedParams.copyright || false;
+        if(parsedParams.copyright) {
+            obj.NSHumanReadableCopyright = parsedParams.copyright;
+        }
 
         return _.merge(obj, custom);
     },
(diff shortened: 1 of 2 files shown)


bastimeyer:
I actually partly reverted my previous changes of f163a07 / 87ee185 in #143. `NSHumanReadableCopyright` was set to `false` before, so I thought it was right the way it was. Should have had a deeper look into the specs :disappointed:... Nice find though :+1:

trevorah:
No worries.

adam-lynch:
This fixed #141 right?

trevorah:
Yep.

PR #156, opened 2015-01-14
fixed downloads after move to nwjs.io

trevorah:
http://node-webkit.org has been moved to http://nwjs.io

Note that this will fix all downloads except for 0.12.0-alpha1 due to the name change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 > Lets you build your [node-webkit](https://github.com/rogerwang/node-webkit) apps for mac, win and linux via cli. It will download the prebuilt binaries for a newest version, unpacks it, creates a release folder, create the app.nw file for a specified directory and copies the app.nw file where it belongs.
 
+## Warning: NW.js 0.12.0-alpha1 is currently unsupported
+node-webkit was [recently renamed](https://groups.google.com/forum/#!msg/nwjs-general/V1FhvfaFIzQ/720xKVd0jNkJ) to NW.js and we are currently working to support it. Follow [#157](https://github.com/mllrsohn/node-webkit-builder/issues/157) for updates.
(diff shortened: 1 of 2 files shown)


trevorah:
In the meantime, a temporary fix is to run the builder with the `downloadUrl` option set to `http://dl.nwjs.io/`.

trevorah:
Looks like http://dl.node-webkit.org is back up and serving a 301. Panic over, but we should fix this at some point.

steffenmllr:
@trevorah if you have a time to come up with a solution I'll add you to the repo and npm - I don't really have time at the moment

trevorah:
I can take a look this evening (in 7 hours time) as it is more than just a one line fix to support 0.12.0 and requires some new tests. If you or any of the maintainers arent around at that time then add me in and i can do the release.

In the meantime, I can add a note to the readme about not currently supporting 0.12.0 in this pull request, and that should fix things for 90% of users.

steffenmllr:
thanks for your help - you have admin rights. whats your npm handle so you can push our a new release?

trevorah:
[trevorah](https://www.npmjs.com/~trevorah)

steffenmllr:
thanks again for helping out, gave you npm rights for this and the grunt module

trevorah:
Cool, no problem.

steffenmllr:
//cc @adam-lynch

trevorah:
v1.0.5 is been published but the grunt module shouldnt need a release, just a fresh npm install to pick up the latest dependency.

adam-lynch:
Thanks @trevorah for helping. @steffenmllr, I'm craaaazy busy right now sorry.

PR #158, opened 2015-01-14
Updated downloadURL and versionNameTemplate to match new nwjs.io url

CoreyClarkPhD:


trevorah:
Fixed by https://github.com/mllrsohn/node-webkit-builder/commit/b86431a31fe97ba7b85e52308dbe399701aa6002 and https://github.com/mllrsohn/node-webkit-builder/commit/74e54720fa4b146de9ac245d68351d922c0e9de6 which is now live as version 1.0.6.

Thanks @CoreyClarkPhD

PR #159, opened 2015-01-15
Implement URL changes

jeanres:
Patch to support new URLs, I know it has already been done but it seems to have broken the build. 

I amended the tests as well to reflect the url changes


jeanres:
Closing this as I needed to fix the app path as well.

PR #160, opened 2015-01-15
Implement node-webkit url changes

jeanres:
This changes the app name, and the urls, I have also fixed the tests to look for the new urls

### DIFF
--- a/lib/platforms.js
+++ b/lib/platforms.js
@@ -20,14 +20,14 @@ module.exports = {
     osx32: {
         runable: 'node-webkit.app/Contents/MacOS/node-webkit',
         files: {
-            '*': ['node-webkit.app']
+            '*': ['nwjs.app']
         },
         versionNameTemplate: 'v${ version }/nwjs-v${ version }-osx-ia32.zip'
     },
     osx64: {
         runable: 'node-webkit.app/Contents/MacOS/node-webkit',
         files: {
-            '*': ['node-webkit.app']
+            '*': ['nwjs.app']
         },
         versionNameTemplate: 'v${ version }/nwjs-v${ version }-osx-x64.zip'
     },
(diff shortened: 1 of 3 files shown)


avihaymenahem:
Can someone please merge? :)

matteocrippa:
+1 it is needed to merge or atm the stock npm version is not working anymore

lib/platforms.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
             '<=0.9.2': ['nw.exe', 'ffmpegsumo.dll', 'icudt.dll', 'libEGL.dll', 'libGLESv2.dll', 'nw.pak'],
             '>0.9.2': ['nw.exe', 'ffmpegsumo.dll', 'icudtl.dat', 'libEGL.dll', 'libGLESv2.dll', 'nw.pak', 'locales']
         },
-        versionNameTemplate: 'v${ version }/node-webkit-v${ version }-win-ia32.zip'
+        versionNameTemplate: 'v${ version }/nwjs-v${ version }-win-ia32.zip'
```
trevorah: Wont this break because the pre 0.12.0 versions of node webkit are still called `node-webkit` not `nwjs`? Look in http://dl.nwjs.io/v0.11.5/ for example.

trevorah:
If the filename changes you propose are wrong, then this is a dupe of #156.

Also, @matteocrippa, a temporary fix is to run the stock builder with the `downloadUrl` option set to `http://dl.nwjs.io/`.

jeanres:
@trevorah if you look @ #156 it only changes the download url not the filenames. So it will fail

trevorah:
It would fail for the latest alpha, but your changes would break for all other versions as those filenames have not changed. If you can get both to work, then that would be excellent!

jeanres:
@trevorah if you are so critical about this, why don't you patch it? I currently don't have the time to patch this

matteocrippa:
@trevorah thanks for the hint!

trevorah:
@matteocrippa no problem.

@jeanres I'll be working on a fix for 0.12.0 this evening (in 7 hours time). Let me know if youre free.

trevorah:
Fixed by https://github.com/mllrsohn/node-webkit-builder/commit/b86431a31fe97ba7b85e52308dbe399701aa6002 and https://github.com/mllrsohn/node-webkit-builder/commit/74e54720fa4b146de9ac245d68351d922c0e9de6 which is now live as version 1.0.6.

Thanks @jeanres

PR #163, opened 2015-01-17
fixed build fail in example. downloading nwbuild version is old.

sassy:
build fail in example.  error message is "Error: unknown platform win32".
This error is caused by older version of node-webkit-builder.
so fixed package.json.


reorx:
This one should be merged LGTM :)

adam-lynch:
Thanks very much

PR #170, opened 2015-01-26
Fix EMFILE errors

felixSchl:
I was still receiving `EMFILE` errors after all and had a closer look at the copy code.

The problem is that the current implementation will launch all copy calls at once, essentially opening possibly thousands of read- and write-streams at the same time.

This patch introduces a method called `mapChunked` that allows us to chunk up the number of parallel copy processes by setting the `chunksize` to something reasonable. This number is driven by `options.maxOpenFiles` but we can rename this if required. I chose not to do any fancy figuring out how many parallel opened files are ok per system code as I feel this is something that a client to the library should be doing.

So far, I have applied the technique only to `mergeAppFiles` but if this patch is approved as a reasonable way forward, it would likely make sense to apply the function to all places where a lot of files are processed at once.

*Note*: I used [this gist](https://gist.github.com/dannycroft/7647231) to save myself the time backporting lodash's `chunk` method to 2.x. Finally, here ia a [proof of concept](https://gist.github.com/felixSchl/8955008908385033e369) of the employed approach.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -38,7 +38,8 @@ function NwBuilder(options) {
         macZip: false,
         macPlist: false,
         winIco: null,
-        argv: process.argv.slice(2)
+        argv: process.argv.slice(2),
+        maxOpenFiles: 128
     };
     // Intercept the platforms and check for the legacy platforms of 'osx' and 'win' and
     // replace with 'osx32', 'osx64', and 'win32', 'win64' respectively.
@@ -430,16 +431,24 @@ NwBuilder.prototype.mergeAppFiles = function () {
 
             // no zip, copy the files
             if(!self.options.macZip) {
-                self._files.forEach(function (file) {
-                    var dest = path.resolve(self.getResourcesDirectoryPath(platform), 'app.nw', file.dest);
+                copiedFiles = Utils.mapChunked(
+                    self._files,
+                    self.options.maxOpenFiles,
+                    function(file) {
+                        var dest = path.resolve(
+                            self.getResourcesDirectoryPath(platform),
+                            'app.nw',
+                            file.dest
+                        );
 
-                    if(file.dest === 'package.json' && platform.platformSpecificManifest){
-                        copiedFiles.push(self.writePlatformSpecificManifest(platform, dest));
+                        if(file.dest === 'package.json' && platform.platformSpecificManifest) {
+                            return self.writePlatformSpecificManifest(platform, dest);
+                        }
+                        else {
+                            return Utils.copyFile(file.src, dest, self);
+                        }
                     }
-                    else {
-                        copiedFiles.push(Utils.copyFile(file.src, dest, self));
-                    }
-                });
+                );
             } else {
                 // zip just copy the app.nw
                 copiedFiles.push(Utils.copyFile(
--- a/lib/index.js
+++ b/lib/index.js
@@ -422,52 +422,52 @@ NwBuilder.prototype.zipAppFiles = function () {
 };
 
 NwBuilder.prototype.mergeAppFiles = function () {
-    var self = this,
-        copiedFiles = [];
+    var self = this;
 
-    this._forEachPlatform(function (name, platform) {
-        // We copy the app files if we are on mac and don't force zip
-        if(name === 'osx32' || name === 'osx64') {
+    return Promise.all(_.foldl(
+        this._platforms,
+        function(acc, platform, name) {
+            var next = null;
+            if(name === 'osx32' || name === 'osx64') {
+                if(!self.options.macZip) {
+                    next = Utils.mapChunked(
+                        self._files,
+                        self.options.maxOpenFiles,
+                        function(file) {
+                            var dest = path.resolve(
+                                self.getResourcesDirectoryPath(platform),
+                                'app.nw',
+                                file.dest
+                            );
 
-            // no zip, copy the files
-            if(!self.options.macZip) {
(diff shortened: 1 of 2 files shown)


adam-lynch:
> The problem is that the current implementation will launch all copy calls at once, essentially opening possibly thousands of read- and write-streams at the same time.

What's wrong with that if graceful-fs is used?

---

`maxOpenFiles` is like something ncp would do and I don't think I like it. I prefer what graceful-fs claims to be, in that if the limit is reached, it's retried until possible. 99% of users wouldn't a clue what to be in as `maxOpenFiles`. Also, of course this could slow down builds for users who's system can handle a lot of open files if they just used the default. Correct me if I'm wrong. 

Then again, if we can't find another way to fix it we might just have to but I still think it *should* be possible for everything rely on graceful-fs underneath and never throw `EMFILE`s.

adam-lynch:
I tried it with my project and got:

>TypeError: Object [object Promise] has no method 'push'
at C:\project-name\node_modules\node-webkit-builder\lib\index.js:462:25
at C:\project-name\node_modules\node-webkit-builder\lib\index.js:592:16
at forOwn (C:\project-name\node_modules\lodash\dist\lodash.js:2105:15)
at Function.forEach (C:\project-name\node_modules\lodash\dist\lodash.js:3302:9)
at NwBuilder._forEachPlatform (C:\project-name\node_modules\node-webkit-builder\lib\index.js:591:7)
at NwBuilder.mergeAppFiles (C:\project-name\node_modules\node-webkit-builder\lib\index.js:428:10)
at tryCatch1 (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\util.js:63:19)
at Promise$_callHandler [as _callHandler] (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\promise.js:695:13)
at Promise$_settlePromiseFromHandler [as _settlePromiseFromHandler] (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\promise.js:711:18)
at Promise$_settlePromiseAt [as _settlePromiseAt] (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\promise.js:868:14)
at Promise$_settlePromises [as _settlePromises] (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\promise.js:1006:14)
at Async$_consumeFunctionBuffer [as _consumeFunctionBuffer] (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\async.js:74:12)
at Async$consumeFunctionBuffer (C:\project-name\node_modules\node-webkit-builder\node_modules\bluebird\js\main\async.js:37:14)
at process._tickCallback (node.js:442:13)

felixSchl:
> What's wrong with that if graceful-fs is used?

Well, it throws `EMFILE` errors on osx, i.e. it does not work as advertised. I would not have bothered if it just worked.

As to the whole graceful approach, it's confusing. We require both `graceful-ncp` and `graceful-fs-extra`. `graceful-fs-extra` uses `graceful-fs` and `ncp`. We use both `graceful-ncp` and `graceful-fs-extra` to do the copying (see `downloader.js` and `utils.js` respectively).

But `graceful-fs-extra` uses `fs-extra` and injects `graceful-ncp` as the dependency. `fs-extra` defines a copy function in terms of `fs.createReadStream` which is not handled by `graceful-fs` (please do correct me if I am wrong here).

I do agree that this function could possibly live in another project however, for the sake of getting this project to work now, this is my proposed solution.

> I tried it with my project and got: [errors...]

I see the problem. It's iterating over both platforms. I must admit I have not tested that and it needs to be fixed. `copiedFiles` is assigned a promise by way of calling `mapChunked`. Removing `Promise.all(...)` from [here](https://github.com/felixSchl/node-webkit-builder/blob/fix-emfile-errors/lib/utils.js#L267) and concatenating, rather than assigning to the `copiedFiles` array should do the trick.

felixSchl:
> I tried it with my project and got: [errors...]

Fixed, but untested under windows. Seems to work fine under mac, but tested only without `macZip` true.

felixSchl:
Let me add that I agree that it would be nice to have a abstraction over performing this IO-heavy work that would ensure the os can keep up and queue things nicely. But I am not sure about the implicit nature of `graceful-fs` aside from the fact that it does not appear to work. I am open to suggestions. We are using this fork in production to get around the `EMFILE` errors`.

felixSchl:
Well, I got fed up with the state of copying files and folders on npm, I wrote [my own library](https://www.npmjs.com/package/kopeer) and use it on the master branch of my fork and everything works fine. `Utils.copyFile` is now one line and calls out to `kopeer` to do the heavy lifting. You should like it, it's heavily based on promises, i.e. monadic design and stateless operation (...IO aside). I have read through both ncp and wrench.js and they both suffer (imo) from callback hell and inherent mess. Although wrench.js does a significantly better job and the code is a **lot** cleaner than ncp.

shennan:
+1 for a merge that resolves copying files/directories with Kopeer (or anything other than what it's using currently). I've just seen it work (albeit remotely) on Windows. It solved an `EPERM 50` error that was being thrown during an osx32 build: trouble with copying icns files. Kooper did the trick. Hoping the PR will pass the test so I can stop patching my/your node-webkit-builder. :-)

adam-lynch:
Going to close as I've not seen an EMFILE in a *long* time or a complaint about it. Feel free to shout at me if I'm wrong.

PR #172, opened 2015-01-27
Do not expand `files`, add `fileglob`

### LINKED ISSUE #171: NwBuilder.options.files needs rename
The `NwBuilder` does not use the `options.files` verbatim, it runs it through a `Glob`. I ran into a case were I built my own grunt wrapper and grunt would already do the globbing for me and pass the desired files into nodewebkit.

I think we should either:

* Not glob
* Call `options.files` -> `options.fileglob`
* Opt-out via flag
* Only glob if filename is a glob pattern, i.e. has leading bang `!` or contains astrix characters `*`.

A rename would probably be the easiest way forward.

Why is this an issue? While the code - as it stands - can be considered "semantically correct", globbing a filepath with a thousand patterns (previously expanded filenames) is extremely slow.

felixSchl:
The first argument to `utils.getFileList` is now a list of already
expanded files, the second argument is a list of fileglob patterns.

Boths lists are concatenated after expansion and made unique.

This allows a user to specify a set of known files or have node webkit
builder expand it on the fly. This is particularily interesting for
grunt, which does the file expansion already.

This changes the `NwBuilder` interface with a backwards incompatible
manner: The constructor now takes both a `files` and a `fileglob`
array. However, what is now called `fileglob` was previously called
`files` and will *NOT* be expanded further.

This is one possible implementation to resolve #171

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -22,7 +22,8 @@ module.exports = NwBuilder;
 function NwBuilder(options) {
     var self = this;
     var defaults = {
-        files: null,
+        files: [],
+        fileglob: null,
         appName: false,
         appVersion: false,
         platforms: ['osx32', 'osx64', 'win32', 'win64'],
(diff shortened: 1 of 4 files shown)


adam-lynch:
See https://github.com/nwjs/nw-builder/issues/171#issuecomment-73319969

PR #184, opened 2015-02-10
Update index.js

waltfy:
Issues:
- Whenever you try to run `Builder.run()`, a folder named `--enable-logging` is created in the application source directory.
- `concat` results in the same output as above but with a `run` folder instead.


adam-lynch:
Sorry for taking so long to get to this. I can't say I've ever seen that happen. Can you clarify?

PR #185, opened 2015-02-12
Pass all kinds of information back from the build cmd

chrisronline:
Ref: https://github.com/mllrsohn/node-webkit-builder/issues/181


adam-lynch:
Thanks!

PR #189, opened 2015-02-17
resolves #188

### LINKED ISSUE #188: Broken on Mac because of \r instead of \n
You're using the carriage return character (used on windows) and not the newline character for line breaks. => It's broken on mac and linux.

```
$ nwbuild -r ./app

env: node\r: No such file or directory
```

konsumer:
resolves #188 and adds nice build commands:

- `npm run publish` - fix lines and publish to npm
- `npm run eol` - just fix lines

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "node-webkit-builder",
   "main": "index.js",
   "scripts": {
-    "test": "colortape test/*.js"
+    "test": "colortape test/*.js",
+    "eol": "node -e \"fs=require('fs'); eol=require('eol'); fs.writeFileSync('bin/nwbuild', eol.lf(fs.readFileSync('bin/nwbuild').toString()))\"",
+    "publish": "npm run eol && mversion patch -m && npm publish"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


adam-lynch:
If you add a custom `install` script and then run `npm install`, it runs the custom one (`npm run install`). I hope that doesn't happen in this case?

adam-lynch:
Actually could you remove the custom `publish` alltogether? I think we'll leave it up to us to the version bump (let them decide between patch, minor & major), running `npm run eol` and `publish`.

Also, could you create a script for the eol stuff instead of putting it inline? Throw it in a `build` directory at the root.

Thanks.

konsumer:
This is totally not my fight. Feel free to reject this PR, and set it up however you want it.

skibz:
@konsumer , i think i speak for a lot of people when i say that your pull request is very much needed and we would __all__ greatly appreciate your addition.

this *is* your fight.

adam-lynch:
I'll look into this tonight if I get a chance

konsumer:
It's an easy fix to set this up however you like.  Here is the thing, for me: I feel like I am being run ragged by bikeshedded contributions, lately. It's been making me super-grumpy. Like I will use a library, maybe for one little thing, even just evaluate it and end up using something else, and it has some glaring thing missing or broken, which through an overactive sense of craftsmanship I can't let go. I submit a PR to fix whatever trivial little show-stopping thing it is, and people are like "oh could you use 4 spaces instead of 2" or "we do that one line there like X". Maybe I just need to accept that this work goes this way, but I don't have a desperate need for this particular project to function, in the first place. If the maintainers don't value the contribution, or don't have the time to checkout and "fix" their own repo (add 2 more spaces, put that line in another file or whatever) it's really not clear why it's now my job, as a non-user of this tool to "fix". I could instead be fixing more actual issues in tools I am using, writing tools that help people get their work done, and have my own libraries & real projects to maintain & build. If you can use the lil code bits I contributed in some way to accomplish your awesome, I couldn't be more pleased. If not, feel free to ignore them.

Maybe I just need a vacation or something, but at least at this moment, that's why this really isn't my fight.

skibz:
fair enough, man.
i know that feeling all too well.

adam-lynch:
Fixed in 1.0.9. 

@konsumer thanks for the help :+1:

konsumer:
No prob. Sorry for the grumpiness.

adam-lynch:
@konsumer nah, you're fine :smile:. I've felt like that too.

PR #192, opened 2015-02-22
Safely access child process reference created when using run()

kessler:
Sometimes you need access to the reference of the child process created when launching nw using run(). This commit attempts to expose it in an orderly manner


adam-lynch:
`1.1.0` published

PR #193, opened 2015-02-23
Add option to disable merging the zip file with the node webkit executable

KalleV:
Optional zip file merging on Windows/Linux is helpful for Node Webkit app updaters that update the source file package separately from the executable.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -37,6 +37,7 @@ function NwBuilder(options) {
         macIcns: false,
         macZip: false,
         macPlist: false,
+        mergeZip: true,
         winIco: null,
         argv: process.argv.slice(2)
     };
@@ -448,6 +449,13 @@ NwBuilder.prototype.mergeAppFiles = function () {
                     self
                 ));
             }
+        } else if (!self.options.mergeZip) {
+            // copy the zipped package.nw into the app directory
+            copiedFiles.push(Utils.copyFile(
+                self.getZipFile(name),
+                path.resolve(platform.releasePath, 'package.nw'),
+                self
+            ));
         } else {
             // We cat the app.nw file into the .exe / nw
             copiedFiles.push(Utils.mergeFiles(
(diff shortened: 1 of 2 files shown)


mfkp:
+1 on this, I need this flag as well. Just tried it out and it works great. Would like to see this merged.

Thanks @KalleV

matthew-dean:
Does this also prevent zipping of the app? Or just prevent concatenation of the zipped file with the .exe? (Would be great to be able to do both.)

williamoliveira:
:+1:

iofjuupasli:
@adam-lynch merge please. Change is very simple, and doesn't break anything. And it works as already said

lib/index.js
```diff
@@ -37,6 +37,7 @@ function NwBuilder(options) {
         macIcns: false,
         macZip: false,
         macPlist: false,
+        mergeZip: true,
```
adam-lynch: It should default to `false` for backwards-compatible, right?

matthew-dean: Isn't zipping on those other platforms defaulted to true? Right now, there is no option for Win/Linux, so it always zips. Hence: a default true value.

adam-lynch: Ah ok, I didn't know it did. Will give this a test soon.

adam-lynch:
Tested on my Mac just to be safe. Will test tomorrow on a Windows machine.

adam-lynch:
Just tested with a real app on Windows. I assume these results are expected, but could someone confirm?

| Project   | mergeZip  | ZIP size | .exe size |
| ------------- | ------------- | ------------- | ------------- |
| nwjs/nw-builder  | N/A  | 43.98 MB | 53,911 KB |
| kallev/node-webkit-builder  | undefined / true | 36.81 MB | 53,911 KB |
| kallev/node-webkit-builder  | false | 36.81 MB | 45,296 KB |

The ZIP mentioned here contains all of the outputted files.

P.S. For my app, there's no noticeable difference in startup time.

adam-lynch:
You need to pull from master here before we can merge this. Also, could we maybe come up with a better name than `mergeZip`?

DblK:
Maybe something like `mergeExecutable` instead of `mergeZip`.
Once this change will be made and the conflict resolved, it will be merged.

ayushmxn:
@KalleV New maintainer here 👋
I know it's been a while but would you like to give this a shot?

ayushmxn:
Wow, this PR has been open for ~7 years. I'll just make the relevant changes myself and probably release this next week.

PR #213, opened 2015-04-09
Copy locales when building for Linux

senshu:
I have also modified the Windows settings: for all platforms, the "locales" folder did not exist before version 0.10.2.


adam-lynch:
We need to merge this and #220

PR #215, opened 2015-04-10
Fixing __MACOSX edge-case

MichielDeMey:
This fixes the edge-case of having a __MACOSX folder inside the zip, which will break the logic defined on line 172 since the rootFiles array will have a length of 2 instead of 1.

Fixes https://github.com/Soundnode/soundnode-app/pull/228

### DIFF
--- a/lib/downloader.js
+++ b/lib/downloader.js
@@ -154,6 +154,14 @@ module.exports = {
         return done.promise;
     },
     stripRootFolder: function(extracted){
+        
+        // Strip __MACOSX folder from rootFiles
+        var rootFolders = fs.readdirSync(extracted.destination);
+        for (var i = 0; i < rootFolders.length; i++) {
+          var folder = rootFolders[i];
+          if (folder === '__MACOSX') rimraf.sync(extracted.destination + path.sep + folder);
+        };
+        
         var done = Promise.defer(),
             files = extracted.files,
             destination = extracted.destination,


adam-lynch:
Why would there be "a  __MACOSX folder inside the zip" though?

MichielDeMey:
Because it was zipped using a Mac. :)
http://superuser.com/questions/104500/what-is-macosx-folder

adam-lynch:
Manually? I.e. not with node-webkit-builder?

MichielDeMey:
Yes, in case of a custom build of node-webkit that you can set using the `downloadUrl` option.

adam-lynch:
Ah ok, cool. I'm on my phone so I'm not sure, is there existing tests for this code? Would it be possible to test this?

MichielDeMey:
You can test it with Soundnode's custom build by pointing the `downloadUrl` to `http://www.soundnodeapp.com/build/` it's using version 0.10.4. And only supports 32-bit builds.

adam-lynch:
I meant it would be good if we could add a test to the node-webkit-builder tests; https://github.com/mllrsohn/node-webkit-builder/tree/master/test

MichielDeMey:
In that case I have no idea how to create a test for this particular use-case.
Any suggestions?

adam-lynch:
Not 100% sure either. I guess there could be a ZIP (with a __MACOSX
directory) stored in
https://github.com/mllrsohn/node-webkit-builder/tree/master/test/fixtures,
then the test would run the function (which does the root directory
stripping) or a parent directory, then checks if the right files are used.
Thinking out loud here.


Regards,
Adam Lynch

MichielDeMey:
That's what I was thinking, it's a straight-forward test scenario but it does add some overhead to the repo since we'll have to store the zip in the fixtures folder.

Any volunteers to write the test?
I can take a stab at it when I have some spare time.

Issue #227, opened 2015-05-20
Do I have to downgrade Nwjs or node-webkit-builder in order to fix this error?


elieobeid7:
I get this error

node-webkit-builder ERROR: The version 1.0.11 does not have a corresponding build posted at http://dl.node-webkit.org/. Please choose a version from that list.

In that list I see http://dl.nwjs.io/?C=M;O=A latest version is 0.13.0, Does it mean I have to downgrade NWJS or choose another version of the builder?

I'm using windows 8.1 (virtualbox) to build an app for all platforms.

NetOperatorWibby:
I'm seeing this issue as well (on Mac). And, this issue is because the necessary files for 0.13 are in another folder called `alpha`, instead of just being in the root folder. I'm not doing development on my nw.io project at the moment, so I'll wait until 0.13 comes out of alpha, but it is kinda of annoying.

elieobeid7:
@NetOperatorWibby if you ever find fix it, please let me know, I'll try the alpha soon but ain't optimistic, if it works I'll let you know too.seems we are on our own, tete a tete. 

I've got more errors using nodejs in my days than haskel, perl, lisp and C combined, i wish i was born in the 50s and worked in 70s and 80s and died or retired by now, i don't like working with those new hot and cool OOP languages, I prefer perl and C any day

chevex:
I solved this by just using version 0.12.2 for now.

```javascript
var nw = new gulp.NwBuilder({
  version: '0.12.2',
  files: ['./package.json', './build/**/*'], // use the glob format
  platforms: ['osx32', 'osx64', 'win32', 'win64'],
  buildDir: './release'
});
```

PS - I don't think you really want to work on code in the 70s or 80s :P. There weren't even step-through debuggers back then! Haha

adam-lynch:
Just specify the `version` now until this is fixed. Closing this in favour of #222

elieobeid7:
sorry for the late reply guys, will try it tomorrow and mark it as fixed, thanks for your help

adam-lynch:
@elieobeid7 No problem... I've closed the issue. I've just pushed a fix for the problem now as well. New version coming soon.

wind13:
The version 0.13.0-alpha3 does not have a corresponding build posted at http://dl.nwjs.io/.

I got this error today.

Fasani:
I get this error today.

Fasani:
It works when I do what chevex said:

[...]

PR #228, opened 2015-05-27
Modified the error message when it fails to download a build.

chevex:
I'm getting this error message as per issue #227. I don't know how to fix it at this time but I thought perhaps the URL hadn't been updated from node-webkit.org to nwjs.io. Looking at the code, the URL is correct; only the error message is incorrect so I figured I'd update the error to eliminate some confusion.


adam-lynch:
:+1:

PR #232, opened 2015-06-15
runApp "appstart" event

bastimeyer:
Right now it's impossible to know when the NW.js process is starting after calling `run()`. The promise will be resolved when killing the application, so the promise itself can't be used for getting the NW.js process.
With this change an `appstart` event will be triggered right after spawning the NW.js process. There's also an error callback added, which will reject the promise in case of an error.


bastimeyer:
Any comments on this?

PR #220, opened 2015-04-22
Update platforms.js

freakinside:
In nwjs > 0.12.0 two dll were missing in the final folder. Needed for webgl, etc, in some windows machines.

PR #238, opened 2015-07-10
added pdf.dll to listed files for versions greater than 0.11.5

joshuaterrill:
```pdf.dll``` is now included in the list of libraries which will make it possible for printing to work again.


freakinside:
Already included on #220

PR #245, opened 2015-07-22
put upper bound to semver check for windows

marianoguerra:
This avoids this version matching if evaluated before the next rule
(>=0.12.0) which would end up not including files like pdf.dll

### DIFF
--- a/lib/platforms.js
+++ b/lib/platforms.js
@@ -6,7 +6,7 @@ module.exports = {
         getRunnable: function() { return 'nw.exe'; },
         files: { // First file must be the executable
             '<=0.9.2': ['nw.exe', 'ffmpegsumo.dll', 'icudt.dll', 'libEGL.dll', 'libGLESv2.dll', 'nw.pak'],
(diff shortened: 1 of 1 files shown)


adam-lynch:
This doesn't need an operator? See https://github.com/mllrsohn/nw-builder/pull/245/files#diff-4b53c119f750f41905184bc503e871c6R34 has `||`. There's no 'and'?

marianoguerra:
I copied from here: https://github.com/marianoguerra/nw-builder/blob/fix-semver-matching-windows/lib/platforms.js#L58 and it seems to work

from semver module docs

    Comparators can be joined by whitespace to form a comparator set, which is satisfied by the intersection of all of the comparators it includes.

https://www.npmjs.com/package/semver

adam-lynch:
Thanks for that. Will publish a new release later.

adam-lynch:
nw-builder@2.0.2 published

marianoguerra:
great, thanks!

PR #248, opened 2015-08-06
Fixed some semicolons

maxrimue:
It doesn't have any impact on the functionality, but for cosmetic reasons :smile:


adam-lynch:
:+1:

PR #256, opened 2015-09-23
Merges https://github.com/speedskater/node-webkit-builder

matthew-dean:
Merges zip extension for all platforms into current nw-builder release. (See: https://github.com/nwjs/nw-builder/pull/200)


adam-lynch:
Closing in favour of #193. Thanks though

PR #260, opened 2015-10-02
Added platform check for OSX32 and rejected build operation

ThorstenHans:
If OSX32 is present in the list of platforms, all builds will break. Because of that build will now be rejected if osx32 is given


adam-lynch:
Sorry I missed this until now ( :/ ) but "osx32 support has been discontinued"? What no longer supports it? :smile:

adam-lynch:
This should be fine as of 2.2.1 because we now read http://nwjs.io/versions.json. See #333. Thanks anyway!

PR #270, opened 2015-11-18
Add new option winZip same as macZip to solve startup slowly issue on windows

leecade:


adam-lynch:
Closing in favour of #193. Thanks for the effort though! I promise quicker replies in future. Just need to test #193 (should do that tomorrow).

adam-lynch:
Merged #200 and released as 2.2.0.

PR #277, opened 2015-12-07
Return non-zero exit code on errors

micopiira:


bin/nwbuild
```diff
@@ -94,5 +94,5 @@ np.then(function() {
     process.exit(0);
 }).catch (function(error) {
     if (error) console.error(error);
-    process.exit(0);
+    process.exit(1);
```
adam-lynch: I wonder why `if (error)` is checked here

adam-lynch:
Thanks!

PR #279, opened 2015-12-24
Fatal error: path must be a string

Jetsly:
when build win32 _needsZip is true ,then build osx32 _needsZip is false,but _needsZip is global


adam-lynch:
Sorry there was no reply to this until now. This should be fixed as of #333 (v2.2.1). Thanks for the pull-request!

Issue #275, opened 2015-12-06
Be smarter about available versions (& support alpha builds)


adam-lynch:
Depends on https://github.com/nwjs/nw.js/issues/4037.

This module would then get and validate versions a lot more solidly.

- All versions would be grabbed reliably.
- It would be a lot easier to avoid alpha builds when the version option is defaulted to `latest`.
- Changes to the platform ZIPs between versions won't break anything.
- People could use alpha builds if they like.
- If the user tries to build for a version which doesn't support that platform, an error will be thrown.

Fixes #239, #253, #254, (probably) #255, #259, #262, and probably more.

gaving:
A big :+1: for this from somebody constantly working off alpha builds!

mib4fun:
update ?  nwjs/nw.js#4037 is kinda fix now, right ?

adam-lynch:
Closed as of 2.2.1 (#337)

Issue #254, opened 2015-09-08
0.13.0-alpha3 linux64 build fails


DamonGant:
Building "latest" on linux64 fails because the archive does not contain a "nw.pak"

c0n5pir4cy:
They've changed the nw.pak in the archive to nw_100_percent.pak; renaming that to nw.pak brings up a bunch of other errors about missing static libraries.

Looks like nw.js might be moving to a structure where all of the resources are in the pak files in the new version. Right now I've just forced the use of an older version.

bmsrox:
How did you solve it ?

c0n5pir4cy:
Hey @bmsrox 

I didn't really fix it, I just passed in the version option to use an older version of nw.js for my linux build option.

Belenn88:
Hey @c0n5pir4cy  which version did you use?
my config:
var nw = new NwBuilder({
        version: ' ? ',
        files: './Player/**',
        platforms: ['win64'],
        currentPlatform: 'linux64',
        winIco: './icons/icon.ico'
    });
Thank you!

c0n5pir4cy:
Hey @Belenn88 

I can't remember what I used as it was a while ago, but try '0.12.3' or 'v0.12.3' as that is the latest version which seems to use an ordinary structure.

Belenn88:
Thank you! @c0n5pir4cy

adam-lynch:
See #275. Any help would appreciated.

petarov:
Bump! Linux builds still fail with:

`
Fatal error: ENOENT: no such file or directory, lstat '/<path>/cache/0.13.4/linux64/nw.pak'
`

I tested 0.13, 0.14, 0.15. Only 0.12 seems to work.

petarov:
@evshiron ah, I see. Thanks for the info 🍺

adam-lynch:
Closing in favour of #329. Thanks for reporting.

PR #281, opened 2016-01-04
v1.13.* files for Linux and Windows

tomlandia:
Makes nw-builder work with v1.13 on Windows provided zip option is set to false. I wonder if the app payload is still supposed to be concatenated to nw.exe/nw from v0.13.

Addresses #254 and #275.


chino23:
@adam-lynch is this going to be added? I'd love to work with the betas now and the ffmpegsumo.dll isn't workign

marianoguerra:
nw.pak must be removed on linux32.files and linux64.files on >=0.13.0-alpha3 as well

adam-lynch:

PR #298, opened 2016-03-25
support NW.js 0.13

bluelovers:


jontro:
Hello, check discussion in this pull request https://github.com/nwjs/nw-builder/pull/296

Dropped support for osx32 bit and osx icons needs to be resolved as well

adam-lynch:

PR #296, opened 2016-03-12
Nwjs 13 support

jontro:
Hello, here is the code for what I have been using for successfully building nwjs 13-rc2.

I have only been able to test mac and windows versions.

PR #301, opened 2016-03-30
Update builder to work with nw13 on windows

andresiraola:


tomlandia:
Redundant with and less complete than with #281 and #296

PR #311, opened 2016-04-25
add currentPlatform option

### LINKED ISSUE #310: don't force platforms
I'm surprised to see you've forced platforms in [/lib/index.js line 584](https://github.com/nwjs/nw-builder/blob/master/lib/index.js#L584):

    var platform = this._platforms[this.options.currentPlatform];

***

`this.options.currentPlatform` is set to the currentPlatform, and it prevents using a 32bit executable on 64bit osx or windows, when both support it.

You can atm build what you want, but you can't run it.

What I want to do, is `nw.run()` in my 64bit machine and let it load the 32bit executable, as the `options.platforms` states in the nw config object.

vankasteelj:
allows to run 32bits executables on 64bits machines. optionnal flag.

fix #310

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,7 +27,7 @@ function NwBuilder(options) {
         appName: false,
         appVersion: false,
         platforms: ['osx32', 'osx64', 'win32', 'win64'],
-        currentPlatform: detectCurrentPlatform(),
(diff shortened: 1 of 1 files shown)


adam-lynch:
See https://github.com/nwjs/nw-builder/issues/310#issuecomment-230121056. Thanks anyway!

PR #313, opened 2016-04-27
Fix for packaging with nwjs >0.12.3

ramiz4:


AzazelN28:
This doesn't work from linux :(

evshiron:
@AzazelN28 Greetings. You may want to give https://www.npmjs.com/package/nwjs-builder a try :)

adam-lynch:

PR #322, opened 2016-05-19
Fixed platforms.js for newer nwjs versions (greater than 0.12.3)

AzazelN28:


adam-lynch:

PR #326, opened 2016-06-03
Added support for newest version of nwjs

joshterrill:
I'll go back through and add the rest of the major versions later, but at least this supports `0.15.1`.


adam-lynch:

joshterrill:
Yup, totally on board with the new plan.

PR #331, opened 2016-06-14
Fix #329

### LINKED ISSUE #329: New versions are not supported
We are focusing on this now; see #327.

DblK:
Add missing windows files for v0.13.0 and up


linquize:
Could you also add new files for v0.13+ for linux32 and linux64?

adam-lynch:
We're not using this for newer versions at all. We're just copying everything over. See #333. Thanks anwyay

PR #333, opened 2016-06-17
Supporting newer NW.js versions

adam-lynch:
For #329.

I have only tested this on Mac OS X as I'm abroad right now and I only have my laptop. I'd appreciate it if someone could run our tests and do some manual tests on an example app with varying config.

~~On Mac OS X, there's a problem when building a Mac app with a newer version (e.g. 0.15.2) in that the app name displayed in Finder is `nwjs.app`. I'm not sure why. I've created a StackOverflow question; [App name in Finder not matching app name in Terminal](http://stackoverflow.com/questions/37890440/app-name-in-finder-not-matching-app-name-in-terminal).~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ cache
 tmp/
 test/temp/
 .DS_Store
+
+example/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - "0.11"
-  - "0.10"
+  - "4.0"
+  - "6.0"
 notifications:
   webhooks:
     urls:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,8 @@
 # Test against these versions of Node.js.
 environment:
   matrix:
-    - nodejs_version: "0.10"
-    - nodejs_version: "0.11"
+    - nodejs_version: "4.0"
+    - nodejs_version: "6.0"
 
 # Allow failing jobs for bleeding-edge Node.js versions.
 matrix:
--- a/example/Gulpfile.js
+++ b/example/Gulpfile.js
@@ -1,16 +1,15 @@
-var NwBuilder = require('../');
+var NwBuilder = require('nw-builder');
 var gulp = require('gulp');
 var gutil = require('gulp-util');
 
 gulp.task('nw', function () {
 
     var nw = new NwBuilder({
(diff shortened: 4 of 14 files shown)


evshiron:
@adam-lynch The display name is not only affected by `Info.plist`, but also `Resources/*.lproj`, which are for i18n purpose. Removing them additionally should be able to bring the correct display name back.

linquize:
Please update `lib/platforms.js` to support new filenames in newer nw.js

adam-lynch:
Thanks @evshiron, that fixed it.

@linquize that's not used for newer NW.js. All files are copied. Did you test this?

linquize:
`bin/nwbuild -p linux64 -o out app`

nw-builder/node_modules/update-notifier/index.js:27
                throw new Error('pkg.name and pkg.version required');

linquize:
Also, the SDK flavour is downloaded. I expect normal flavour is downloaded.

linquize:
The zipped content is written to `chromedriver.exe`. Shouldn't be `nw.exe`?

linquize:
Travis-CI node.js version 0.10 is too old. I am using node 4.x or 6.x

linquize:
Travis-CI: You need to upgrade Travis-CI node.js version to 4.x or 6.x.
There is an error with old node.js 0.10 when running test
Unhandled rejection TypeError: Object [object Object],[object Object] has no method 'find'

adam-lynch:
@linquize Hmm I didn't get any error when building for Linux 32 & 64 on Mac. Will test on Windows once I update my Node version there.

>Also, the SDK flavour is downloaded. I expect normal flavour is downloaded.

I think for the first patch version I'll use the SDK flavour so it's more inline with how older versions worked. Then I'll add the ability to select flavours after that.

>The zipped content is written to chromedriver. Shouldn't be nw?

Oh nice catch. I was worried about one line actually, I think that's it. Will look into it.

>You need to upgrade Travis-CI node.js version

Thanks.

adam-lynch:
@linquize fixed the chromedriver issue and the fact it wasn't renaming the exec to your chosen app name. Give it another go please.

adam-lynch:
I thought the `winIco` option wasn't working for newer versions but I just confirmed it was fine when building on my Mac or a Windows VM on my Mac. Also tested building for Linux on my Mac and building on Linux for Linux. So... merging 😃

PR #336, opened 2016-07-02
Implementation of flavor feature

DblK:
The flavor feature allow you to select the flavor "normal", "sdk" when
using nw-builder.

### DIFF
--- a/lib/Version.js
+++ b/lib/Version.js
@@ -22,7 +22,9 @@ module.exports = function Version(args){
         generatePlatformUrls = function(version, supportedPlatforms){
             var platformUrls = {};
             supportedPlatforms.forEach(function (supportedPlatform) {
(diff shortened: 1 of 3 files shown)


DblK:
It's normal everything failed because I changed the signature of the functions.
It needs to adjust test and also write new tests.

lib/Version.js
```diff
@@ -32,24 +34,26 @@ module.exports = function Version(args){
             return platform.replace('32', '-ia32').replace('64', '-x64');
         };
 
-        var mapPlatformToExtension = function(platform){
+        var mapPlatformToExtension = function(platform, flavor){
```
adam-lynch: This isn't used?

adam-lynch:
Thanks for the pull-request! I published the fix for supporting newer versions today and had to make a couple of fixes. Sorry, do you mind basing this of master instead? 😊

DblK:
Okay, I will try to do it soon enough.

DblK:
See #340

PR #338, opened 2016-07-02
SyntaxError: Unexpected token ILLEGAL

RedSparr0w:


adam-lynch:
Ugh how did that happen... thanks!

PR #340, opened 2016-07-03
Implementation of flavor feature based on new master

DblK:
It implements a new parameter "flavor" which is by default "sdk" that
allow to choose the flavor of the build (useful for dev and production
application).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -33,6 +33,7 @@ function NwBuilder(options) {
         buildDir: './build',
         cacheDir: './cache',
         downloadUrl: 'http://dl.nwjs.io/',
+        flavor: 'sdk',
         buildType: 'default',
         forceDownload: false,
         macCredits: false,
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -185,6 +186,7 @@ NwBuilder.prototype.resolveLatestVersion = function () {
 
 NwBuilder.prototype.checkVersion = function () {
     var version = this.options.version,
+        flavor = semver.satisfies(version, '<0.12.3') ? 'sdk' : this.options.flavor,
```
linquize: Should move this line above `var getVersionFromManifest = function(){`

lib/index.js
```diff
@@ -186,8 +186,8 @@ NwBuilder.prototype.resolveLatestVersion = function () {
 
 NwBuilder.prototype.checkVersion = function () {
     var version = this.options.version,
-        flavor = semver.satisfies(version, '<0.12.3') ? 'sdk' : this.options.flavor,
         self = this;
+    var flavor = semver.satisfies(version, '<0.12.3') ? 'sdk' : this.options.flavor;
```
linquize: OOPS, this is not the right position. 

Should move to the line above `var getVersionFromManifest = function(){`

gaving:
:+1: (wanting this for production builds)

adam-lynch:
Hey, thanks for this! Please add / update the tests though 😀

bhavyaw:
@adam-lynch @DblK Any chance it will get merged ?

Is there any other we can opt for a specific flavor rather default sdk one.

chrisneal:
Any movement on this?

bhavyaw:
I don't know when this will be merged into repo but i was able to figure out which file we need to make the changes so we can download the **normal flavor**.

Here's what you need to do :

1. Open file `Versions.js` ( Capital V ) 
2. Search for method `generatePlatformUrls`
3. Replace String in that method from '/nwjs-sdk-v' to '/nwjs-v'


downloadingsdkflavor.

PR #355, opened 2016-08-19
Download NW.js archives via https

bastimeyer:
Secure downloads should be the default...
https://github.com/nwjs/nw.js/issues/3228#issuecomment-237153818

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -32,7 +32,8 @@ function NwBuilder(options) {
         version: 'latest',
         buildDir: './build',
         cacheDir: './cache',
-        downloadUrl: 'http://dl.nwjs.io/',
+        downloadUrl: 'https://dl.nwjs.io/',
+        manifestUrl: 'https://nwjs.io/versions.json',
         flavor: 'sdk',
         buildType: 'default',
         forceDownload: false,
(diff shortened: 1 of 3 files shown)


adam-lynch:
This should just update to `https`. Addition of the `manifestUrl` option should be another PR. Is it really needed anyway?

bastimeyer:
My idea behind this was enabling a future addition of custom manifestUrls in the builder options, since the downloadUrl can be customized already. Theoretically it is not needed and the https manifest url should be used at all times, but without this change all tests would need to be updated, too, and this was the simpler approach.
I'm fine with closing this PR if you want to change it and upgrade everything by yourself.

DblK:
Since we change the file, could you update your pull request to resolve the conflicts?
After the review, I will merge it.

bastimeyer:
Conflicts solved, rebased and PR updated...
I've also added the `manifestUrl` option. Should I add this to the readme? The `downloadUrl` option is not listed in the readme, so I was not sure.

Btw, the flavors addition which caused these conflicts doesn't seem to have any tests. Also, the `flavor` parameter in `getLegacyVersions`, `getVersionsFromManifest`, `getLatestVersion`, `getVersion` and `getVersions` is never used. Just wanted to mention this...

adam-lynch:
This is slightly related to #364 so `manifestUrl` would be handy, except there is no alternative location where `versions.json` is hosted.

I would've thought we could phase out `downloadUrl` until I saw #364.

Is there a benefit to having two separate properties (`manifestUrl` and `downloadUrl`)? I guess so?

bastimeyer:
> Is there a benefit to having two separate properties

Having one customizable url and another static one doesn't make sense IMO. How would you solve this problem instead? Both options don't share the same domain name.

The `manifestUrl` could be useful for people who have forked nwjs and are using their own builds with nwbuilder. It's also useful for people who are unable to access nwjs.io or who don't want to use https for whatever reason.

bastimeyer:
@adam-lynch @DblK 
Is there something that needs to be changed?

adam-lynch:
Thanks :smile:

PR #388, opened 2016-10-26
Improve docs around run

1j01:
And fix some formatting via invisible end-of-line whitespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ npm install nw-builder -g
 Yes, there is also a [Grunt Plugin](https://github.com/nwjs/grunt-nw-builder). For Gulp, just use the module :)
 
 
-## Usage
+## CLI Usage
 
 ```shell
 Usage: nwbuild [options] [path]
@@ -39,7 +39,7 @@ Options:
 #### Run NW.js
 During development you can run NW.js with `nwbuild -r path/to/your/younwapp/`
 
-Or use the module:
+## Module Usage
 
 ```js
 var NwBuilder = require('nw-builder');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ Options:
 #### Run NW.js
 During development you can run NW.js with `nwbuild -r path/to/your/younwapp/`
 
-Or use the module:
+## Module Usage
```
adam-lynch: "Programmatic" would be a better word to use

README.md
```diff
@@ -80,22 +88,22 @@ The path to your node webkit app. It supports [simple-glob](https://github.com/j
 
 
 #### options.version
-Type: `String`
-Default value: `'latest'`
+Type: `String`
```
adam-lynch: Accidental?

README.md
1j01: Maybe, but this way it's a noun both times before "usage"; spaghetti/linguine imo

README.md
1j01: No, the two spaces indicate a line break in markdown. They were missing from some options which made the Type and Default value show on the same line. The two spaces before an *empty line* don't matter but I added them as opposed to removed them for consistency because I think there were more that had them than didn't.

1j01:
(Also only one of those comments made it into my "review" because I hit "Add single comment" the first time, but the second time I just hit the pretty green button, and then it said it wouldn't keep my comments if I canceled the review but maybe it was just trying to say that it wouldn't keep the duplication of the comments because that's apparently what the end result of the "review" was.)

1j01:
So, can this be merged?

1j01:
...?

jonathantneal:
Merge-able and helpful.

adam-lynch:
Thanks!

PR #389, opened 2016-10-27
fix(platform): add osx multiple locale support

nurishi:
Ran into the issue that OSX multi locale support becomes unavailable when we build with nw-builder. (Tested with nw-builder 3.1.2 and nwjs 0.14.7 SDK build which works fine with their official build. FYI, currently 0.14.7 Normal build has an issue in multi locale support nwjs/nw.js#4824 so you can only check it with SDK build unless you manually apply the fix suggested in the ticket to Normal build)

Found out the root cause of the issue and it was in the part of omitting creation of Contents/Resources/*.lproj directories. I understand that this line was required to support the custom app name but instead of dropping the whole directory, it should only ignore InfoPlist.string file inside so that the directory itself will be still generated.

Hope this pull request could be merged and save the people facing the same issue.


adam-lynch:
Thanks @nurishi. Clear description. I can't test it right now though unfortunately

lib/index.js
```diff
@@ -413,11 +413,11 @@ NwBuilder.prototype.copyNwjs = function () {
                 var options = {};
 
                 if(['nw', 'nwjs.app', 'nw.exe'].indexOf(destFile) !== -1){
-                    // ignore nwjs.app/Contents/Resources/*.lproj, otherwise the app name will show as nwjs.app in
+                    // ignore nwjs.app/Contents/Resources/*.lproj/InfoPlist.strings, otherwise the app name will show as nwjs.app in. *.lproj directory itself needs to be kept to support multiple locales.
```
zaygraveyard: The word `Finder` (from the next line) should come after `nwjs.app in` (as in `... nwjs.app in Finder. *.lproj ...`)

charlielee:
This PR is still required and would fix issue my issue #524

PR #396, opened 2016-11-11
Fix cache check

armybean:
Fixed check of cached files
Only create cache directory and log message if necessary


adam-lynch:
With this change, couldn't the creation of the cache directory fail silently? I'm not sure what this is fixing (I'm not talking about the `flavors` fix here).

PR #435, opened 2017-03-06
osx 0.20.3 normal only has 2 files at root folder

suziwen:
image


adam-lynch:
Thanks! Will be in the next release

PR #403, opened 2016-12-03
feat(cli): improve options

finscn:
This is a update version of #367

This version is clearer.

And the reason about why I create this PR is at  https://github.com/nwjs/nw-builder/pull/367#issuecomment-264659757

PR #439, opened 2017-03-24
remove osx32 from the default list of platforms

preaction:
Since 32-bit OSX isn't supported by NW.js anymore (because OSX hasn't been released for 32-bit platforms since 2011), it's more friendly to new developers to not give them an error using the default `nwbuild` configuration.

This appears that it will conflict with #403, so if that gets merged I can rebase this.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,7 +27,7 @@ function NwBuilder(options) {
         files: null,
         appName: false,
         appVersion: false,
-        platforms: ['osx32', 'osx64', 'win32', 'win64'],
+        platforms: ['osx64', 'win32', 'win64'],
         currentPlatform: detectCurrentPlatform(),
         version: 'latest',
         buildDir: './build',
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Yes, there is also a [Grunt Plugin](https://github.com/nwjs/grunt-nw-builder). F
 Usage: nwbuild [options] [path]
 
 Options:
-  -p, --platforms      Platforms to build, comma-sperated, can be: win32,win64,osx32,osx64,linux32,linux64   ['osx32', 'osx64', 'win32', 'win64']
+  -p, --platforms      Platforms to build, comma-sperated, can be: win32,win64,osx32,osx64,linux32,linux64   ['osx64', 'win32', 'win64']
   -v, --version        The nw version, eg. 0.8.4                                             [default: "latest"]
   -r, --run            Runs NW.js for the current platform                                   [default: false]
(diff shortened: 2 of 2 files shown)


adam-lynch:
@preaction thanks for this. The default is mentioned in the README too. Could you update that as well please?

preaction:
Done

lib/index.js
```diff
@@ -47,10 +47,10 @@ function NwBuilder(options) {
         argv: process.argv.slice(2)
     };
     // Intercept the platforms and check for the legacy platforms of 'osx' and 'win' and
-    // replace with 'osx32', 'osx64', and 'win32', 'win64' respectively.
+    // replace with 'osx64', and 'win32', 'win64' respectively.
     if(typeof options.platforms != 'undefined'){
         if(options.platforms.indexOf('osx') >= 0){
-            options.platforms.splice(options.platforms.indexOf('osx'), 1, 'osx32', 'osx64');
+            options.platforms.splice(options.platforms.indexOf('osx'), 1, 'osx64');
```
adam-lynch: Sorry now, I'm just looking at this again and thinking this is actually a breaking change. I.e. it would break people's builds relying on it. So if it was merged, it would require a major version bump (to 4.0.0), which would then require people to update their package.json / reinstall. I'm not sure this is worth it. Maybe we could not change this part (`osx` -> `osx32`, `osx64`), merge the rest, and wait until a 4.0.0 to include this specific change. What do you think?

preaction: That's wholly up to you. I would agree that it's a breaking change, but in that same respect all of it is: `osx32` used to be built by default, and now it won't be. I'd think that the opt-in to all OSX platforms with `--platforms osx` is more explicit than using the default, and more likely to cause problems with being changed (which I imagine is what you're thinking).

So, I'll remove this bit presently.

adam-lynch: Ugh, you're right about it all being a breaking change :/

I love merging stuff from new contributors but maybe we should wait. We can both keep an eye out for more breaking changes so we can merge this ASAP 😛

preaction: I mean, the reason I would feel confident about making the changes that are left is because NW.js itself has dropped support for the platform: `osx32` cannot work if you upgrade NW.js. If they want nw-builder to work with older versions of NW.js, they can update their build scripts to pass in the right `--platform` flag. But still totally up to you.

adam-lynch: OK. I'm so easily convinced 😁 

> But still totally up to you.

I'll direct any hate your way 😛.

PR #469, opened 2017-08-30
Fixed always detect osx32 when in run mode. And add -n options to set…

bramblex:


bin/nwbuild
```diff
@@ -9,7 +9,6 @@ var argv = optimist
     .usage('Usage: nwbuild [options] [path]')
 
     .alias('p', 'platforms')
-    .default('p', 'osx32,osx64,win32,win64')
```
adam-lynch: Why not just remove `osx32` here?

bin/nwbuild
```diff
@@ -54,9 +54,10 @@ if(!files) {
 }
 
 var options = {
+    appName: argv.name,
```
adam-lynch: Cool, this still falls back to the name in `package.json` if it doesn't exist yeah?

bin/nwbuild
bramblex: i'am not share if someone using the old version node on osx32 system.

bin/nwbuild
bramblex: of course. https://github.com/nwjs-community/nw-builder/blob/master/lib/index.js#L185

bin/nwbuild
adam-lynch: I mean you're moving the default options somewhere else. Wouldn't it work the same if you changed this line to `.default('p', 'osx32,osx64,win32,win64')`?

zaygraveyard: I agree with @adam-lynch, removing `osx32` here should be equivalent (unless I'm missing something).

PS: @adam-lynch you mean `.default('p', 'osx64,win32,win64')`, right?

bin/nwbuild
```diff
@@ -74,6 +75,11 @@ if(argv.r) {
     options.currentPlatform = currentPlatform;
 }
 
+// Set default target platforms here
+// Remove osx32 from default, nwjs didn't support osx32 yet
+if (!options.platforms) {
+    options.platforms = ['osx64', 'win32', 'win63']
```
zaygraveyard: Typo: `win63` should be `win64`

bastimeyer:
@kessler You have merged and published the typo mentioned by @zaygraveyard

kessler:
@bastimeyer thanks for spotting me! fixed

PR #477, opened 2017-10-10
Fix preparePlatformSpecificManifests to handle edge cases (fixes #476)

### LINKED ISSUE #476: Build stops silently when the manifest has platform overrides
When:
* "package.json" contains platform overrides for a platform **_A_** but not **_B_**
* and a build for platform **_B_** is ran

Then the build stops silently after fetching nwjs (or choosing to use the cache).

zaygraveyard:


adam-lynch:
Thanks a lot @zaygraveyard! Would you help this project more? We're always looking for maintainers/contributors. That could be improving documentation, replying to / closing issues, merging pull-requests, fixing bugs, etc... whatever you like 😄

PR #483, opened 2018-02-25
Apply option.winVersionString even if there is no option.winIco

daisykoo:
Thanks for the amazing nw-builder. Found a small problem when using it, if missing option.WinIco, option.winVersionString will not work. I think it should be independent of each other. This is the really first time I make a contribution, so please let me know if I did any thing wrong. Thanks again!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -674,7 +674,7 @@ NwBuilder.prototype.handleWinApp = function () {
         allDone = [];
 
     this._forEachPlatform(function (name, platform) {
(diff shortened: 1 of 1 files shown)


adam-lynch:
Thanks a lot @daisykoo! Would you help this project more? We're always looking for maintainers/contributors. That could be improving documentation, replying to / closing issues, merging pull-requests, fixing bugs, etc... whatever you like 😄

daisykoo:
I would love to~ thanks for the invitation ^-^.

PR #490, opened 2018-05-01
update request module to latest because of CVE-2018-3728

kessler:


kessler:
published to npm v3.5.2

PR #496, opened 2018-07-13
Add optional compile using nwjc

julijane:
These changes add a new configuration parameter "compileFiles" which is a array of globs. Any matching files will be compiled using nwjc during ZIP packing. To be able to do so the sdk is also downloaded (but used only for nwjc) when this option is set, even if the target flavor is 'normal'.

Usage example:

```
  var nw = new NwBuilder({
    files: './dist/**/**',
    platforms: [ 'linux64', 'win64' ],
    version: '0.31.5',
    buildDir: './releases',
    cacheDir: './cache',
    compileFiles: [ 'dist/**/main.js' ],
    flavor: 'normal'
  })

```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -73,6 +73,7 @@ function NwBuilder(options) {
         macPlist: false,
         winVersionString: {},
         winIco: null,
+        compileFiles: [],
         argv: process.argv.slice(2)
     };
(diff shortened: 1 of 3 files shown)


julijane:
closing this for now, will rework the change

PR #497, opened 2018-07-13
Fix resource settings handling, allow to set icon with rcedit

julijane:
The current code regarding setting windows resources on the executable is somewhat broken. It will not work when only winVersionString is set. It only runs when winIco is set, but then it also overwrites the winVersionString even when it is not set in the config.

Also this change adds a new option 'useRcedit' which when set will set the icon with rcedit instead of calling winresourcer. Apparently nw-builder used to use rcedit for the icon until it was changed in 2014 because there were some problems. I don't know if this was fixed in rcedit, but with a properly prepared icon-file I had no problems at all setting it with rcedit, but I could not get winresourcer running on Linux with wine.

So with this flag the user can now choose if he wants/must use winresourcer or if rcedit will do, which is less of a hassle.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -73,6 +73,7 @@ function NwBuilder(options) {
         macPlist: false,
         winVersionString: {},
         winIco: null,
+        useRcedit: false,
         argv: process.argv.slice(2)
     };
     // Intercept the platforms and check for the legacy platforms of 'osx' and 'win' and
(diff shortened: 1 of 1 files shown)


kessler:
dear @julijane, I'm going over the PRs and issues. Could you please update if this is still relevant? If it does I'd love to chat a little about all these windows flags to better understand the problem. Cheers :-)

julijane:
@kessler Well, unless it has been fixed in the meanwhile (I haven't checked but I don't think so), the bug this fixes is still relevant:

> It will not work when only winVersionString is set. It only runs when winIco is set, but then it also overwrites the winVersionString even when it is not set in the config.

The change to allow to use RcEdit instead of WinResourcer is also still relevant as it adds another option which users might find better working (at least it was like that for me).

Both changes should probably be split into separate commits though for more clarity.

kessler:
@julijane thank you for your reply, to be perfectly honest, I need to put some time into understanding this PR and it's motivation. I will try to get to in the next month or two and will update this thread.

PR #520, opened 2019-08-05
chore: support custom nwjs download url

keenghost:
connect is unstable from some region, and there are lots of mirrors that we can use. we can offer a custom way for users.


cs64188:
we really need this feature!

ayushmxn:
@keenghost @cs64188 Closing this for now. Comment here if you still need this.

PR #540, opened 2021-08-06
Update nw.js releases link

pbek:
The wiki page doesn't seem maintained any more and is out of date.


ayushmxn:
@pbek redirected to the tags page since the releases page hasn't been used. I plan to update the wiki but in the mean time thank for this!

ayushmxn (approved):

PR #560, opened 2022-05-20
chore(docs): add code of conduct

TheJaredWilcurt:
Adding the same Code of Conduct used in all other NW Utils repos.

Since you are the primary maintainer of this repo, you should read it and ensure you agree with it before merging. If you would prefer to use an alternative Code of Conduct, that's cool too. We can also talk through any specific details or questions here or on Gitter.


ayushmxn:
LGTM

PR #577, opened 2022-06-15
JSDoc Linting 1 (plugin)

TheJaredWilcurt:
1 of 3 PR's around JSDoc linting.

These first two are different approaches to doing the same thing.

This shows pulling in a dependency that configures the linting rules for you. It is very clean and simple to maintain. However all rules are hidden away. You can of course still override any of these rules as you see fit. However if most will be overridden, it is better to just handle the rules by hand, like in the other PR example ([JSDoc Linting 2](https://github.com/nwutils/nw-builder/pull/578)).


package.json
```diff
@@ -91,6 +91,10 @@
       "es6": true,
       "node": true
     },
+    "extends": [
+      "eslint:recommended",
+      "tjw-jsdoc"
```
ayushmxn, line 96: I realise I didn't enable the jsdoc eslint plugin here. Not too familiar with `tjw-jsdoc`. Any notable differences between `tjw-jsdoc` and `jsdoc` plugin?

ayushmxn:
I looked at the other PRs and identify with this one the most since it's the easiest to maintain. I looked at the recommended rules (enabled via `plugin:jsdoc/recommended`) and this makes sense to me. Since I'm pretty new to JSDoc, I'm open to suggestions!

package.json
TheJaredWilcurt: PR 1 and PR 2 are just about *how* to set up the rules.

PR 3 is about applying those rules to the codebase and updating files. If there is something you wish was handled different with how the JSDoc blocks are formatted, or how strictly enforced they are, you can comment in PR 3 what changes you'd want and I'll try to find the rule/setting to adjust for that. As it is right now, there's a lot of comment blocks that need descriptions added to explain what a function is doing (like 200ish). So it's a lot of work to document every single thing the library does, and can be very tedious. It does make the code base much easier to work in for new people though, you don't have to read a 20 or 80 line function to understand what it does, the comment above it will always tell you what it does in a quick sentence, what it takes in, and what it returns. But again, that's a nice thing for new people, but a ton of work to set up all at once. Going through and just documenting the types of everything will take a while, but will at least give development benefits as a TS engine will highlight things it doesn't like then.

Or you can skip all this, and just use JSDocs to define out the API layer to give types to other repos. This is equivelant to having a `.d.ts` file in the root of the repo, except [JSDocs gives much better results than TypeScript for API interfaces](https://github.com/nwutils/create-desktop-shortcuts/pull/69).

ayushmxn: > This is why I did two PR's comparing. If you 100% agree with me on everything (or don't care), then this is the simpler approach, you offload caring about the settings to someone else.

I'm honestly fine with this.

> However, if there are any parts of it you disagree with, you can simply add that rule into the rules section here and override those specific things, while still inherriting the rest of my rules.

As I learn more about JSDoc, tweaks can be made when required. This looks like a good start!

> The point is that eslint-plugin-jsdoc has a ton of options, but by default almost all of them aren't turned on. This turns on most of them and adjusts their settings to my liking. I just don't want to force my opinions on your repo.

I know that you know more about JSDoc and JavaScript than me and for me it's not about forcing your opinion on the repo but me learning some new. My first coding gig involved TypeScript and I learnt (and am still learning) the quirks of JavaScript after it. I've come to realise that where TypeScript is a better fit and where JavaScript be be better utilised!

> As it is right now, there's a lot of comment blocks that need descriptions added to explain what a function is doing (like 200ish). So it's a lot of work to document every single thing the library does, and can be very tedious.

The approach I was thinking of was adding `eslint-config-tjw-jsdoc` to the eslint config via this PR, turn off linting in ci and gradually add the jsdocs whenever a code change made lacks it. Let me know what you think?

>  It does make the code base much easier to work in for new people though, you don't have to read a 20 or 80 line function to understand what it does, the comment above it will always tell you what it does in a quick sentence, what it takes in, and what it returns.

Completely agree. The number and quality old contributions would definitely go up because of it. It would also be easier to maintain for the next maintainer(s) after me.

ayushmxn:
@TheJaredWilcurt after resolving the merge conflict and removing linting from the ci, this is good to go!

.github/workflows/cd.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: "16"
 
-      - run: npm install
+      - run: npm ci
```
TheJaredWilcurt, line 16: `npm ci` will install the **exact** modules listed in the `package-lock.json` file. So you know that your published version will work exactly the same as it did for you locally. Without a lock file, or using `npm install` (which updates the lock file), you can end up with a sub-dependency at a different version which may cause unforseen issues.

`npm ci` basically solves the "works on my machine" issue. Now it works on all machines.

ayushmxn (approved):
🚢

PR #583, opened 2022-06-27
Update dependencies

TheJaredWilcurt:
* Updated all dependencies
  * Documented the one dependency I couldn't easily update
* Removed all `pnpm` stuff, to simplify the repo.
  * `package-lock.json` is from `npm`
  * docs now give `npm` instructions
  * `npm` scripts no longer use `pnpm`
  * ci/cd uses `npm ci`
* Updated GHA to use the `package-lock.json`


.github/workflows/cd.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: "16"
 
-      - run: npm install
+      - run: npm ci
```
TheJaredWilcurt, line 16: `npm ci` will install the **exact** modules listed in the `package-lock.json` file. So you know that your published version will work exactly the same as it did for you locally. Without a lock file, or using `npm install` (which updates the lock file), you can end up with a sub-dependency at a different version which may cause unforeseen issues.

`npm ci` basically solves the "works on my machine" issue. Now it works on all machines.

package.json
```diff
@@ -19,23 +19,25 @@
...
-    "test": "pnpm test:unit && tape './test/*.cjs'",
+    "test": "npm run test:unit && tape './test/*.cjs'",
     "test:unit": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
-    "demo": "cd ./test/demo && pnpm start",
+    "demo": "cd ./test/demo && npm start",
     "build": "esbuild src/index.js --bundle --minify --platform=node --outfile=./dist/index.cjs"
   },
+  "ManifestComments": [
+    "Pinned update-notifier to 5.1.0, v6+ requires ESM"
+  ],
```
TheJaredWilcurt, line 29: Unfortunately JSON files do not support comments. So I've added this section here, so documentation around the `package.json` file can still occur.

If you think `package.json` files should be more permissive to allow for comments, you can +1 this post:

* https://github.com/npm/rfcs/issues/291

package.json
```diff
@@ -19,48 +19,50 @@
...
+    "plist": "^3.0.5",
+    "progress": "^2.0.3",
+    "rcedit": "^3.0.1",
+    "recursive-readdir-sync": "^1.0.6",
+    "request": "^2.88.2",
+    "rimraf": "^3.0.2",
+    "semver": "^7.3.7",
+    "simple-glob": "^0.2.0",
+    "tar-fs": "^2.1.1",
+    "temp": "^0.9.4",
```
TheJaredWilcurt, line 61: Not sure what the story is with this. My assumption is that @adam-lynch forked `temp`, made changes and submitted those as a PR. Then didn't want to wait for them to be merged, so pointed to his own fork. It looks like there is no code difference between his version and the official one now, except there have been many commits/new releases done on the official `temp`. I could be wrong on that. But after updating it, I didn't see any red flags go off (I ran all the npm scripts, and none failed).

adam-lynch: Sounds fine to me

.github/workflows/ci.yml
```diff
@@ -12,15 +12,15 @@ jobs:
...
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: volta-cli/action@v1
         with:
-          node-version: ${{ matrix.node }}
-
-      - run: npm run install
+          node-version: ${{ matrix.node-version }}
+          npm-version: '8.5.0'
+      - run: npm ci
```
TheJaredWilcurt, line 23: **Changes:**

* `node` => `node-version`, most repos name the node matrix this, don't know why, but just making it like the rest.
* `npm 8.5.0` - By default Node 14 comes with npm 6, which supports v1 lock files. But this PR includes a V2 lockfile. So we need to use npm 7 or above for that.
* `actions-setup-node` => `volta-cli/action`, This lets us easily set the Node and npm versions in one go, so we can do Node 14, 16, or 18 all with npm 8.5.0
* `npm run install` => `npm ci`, this makes sure we download the exact same dependency versions that were used during development

PR #585, opened 2022-06-30
try running nwjs for mac silicon

ayushmxn:


ayushmxn:
To test for M1s need to add [self hosted runner](https://github.com/github/roadmap/issues/528). Currently only supports [Mac 10, 11 and 12](https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job).

Issue #595, opened 2022-07-15
Regex matching package.json is wrong


davedoesdev:
### Issue Type

- [X] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour

Matches `package-json` package and thinks it's a `package.json` file, causing

```
Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (node:fs:739:3)
    at tryReadSync (node:fs:439:20)
    at Object.readFileSync (node:fs:485:19)
    at /home/david/browser-http2-duplex/node_modules/nw-builder/dist/index.cjs:6:18526
```

### Expected/Proposed Behavior

Shouldn't regard `package-json` directory as `package.json` file

### Additional Info

`package-json` is a published package on npm: https://www.npmjs.com/package/package-json

- Operating System:

Ubuntu 22.04

- Node version:

v18.2.0

- NW.js version:

3.8.0
- Repro link:
- ...

The regex here is wrong: 

* https://github.com/nwutils/nw-builder/blob/f5daaabef4446842f714748a205788bd1269d039/src/utilities/checkPkgOptions.js#L17
* https://github.com/nwutils/nw-builder/blob/6282f27e26499ae835450b4f9f6296d0d9071000/lib/utils.cjs#L54

The period needs to be escaped.

ayushmxn:
Right, that makes sense. Oversight on my part! I didn't write tests for the `checkPkgOptions` function which is probably this is being caught later. Do you want to open a PR? I'll be afk till Monday.

davedoesdev:
Is the intent to match files called `package.json` or files ending with `package.json`?

davedoesdev:
I made a PR assuming only files called `package.json` should be read.

ayushmxn:
Your assumption was correct!

PR #600, opened 2022-07-17
Only read files called package.json #595

davedoesdev:
Fixes: #595

### Changes

- ...


ayushmxn (approved):
🚀

lib/utils.cjs
```diff
@@ -51,7 +51,7 @@ module.exports = {
 
       matches.forEach(function (file) {
```
CosmoMyzrailGorynych, line 52: With `directory/**` glob pattern, it reads **every** package.json and will use any `package.json` file, including those in `node_modules`. And of course it will use the last one, from dependencies.

CosmoMyzrailGorynych (commented):
This 👏 is 👏 broken 👏

lib/utils.cjs
davedoesdev: This PR changed from matching anything ending in `package.json` to anything named `package.json`.
Therefore the number of matches should be reduced.

Globbing was present before.

Issue #689, opened 2022-11-17
Desktop Entry file for linux builds is not configurable


tharatau:
### Issue Type

- [x] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour

- The `.desktop` file for Linux builds lacks a lot of properties
- User cannot update these properties
- Rename `nw` executable to `package.json`'s `name` property.

### Expected/Proposed Behaviour

- Get required properties from `package.json` and provide user with option to override them and provide more properties via the  a `linuxCfg` in the `srcDir/package.json`

### Additional Info

- Operating System: N/A
- Node version: N/A
- NW.js version: N/A
- Repro link: N/A

PR #690, opened 2022-11-17
fix: desktop entry file

tharatau:
Fixes: #689 

- [x] Formatted
- [x] Linted
- [ ] Test coverage

### Changes

- Configure Desktop Entry file via `linuxCfg` property in `package.json` (Consult the [specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html) for more info)
- Rename executable
- Update default cache directory location to user's current working directory
- Log messages to console and save in an `nwbuild.log` file


tharatau:
If a `linuxCfg` property exists in the `srcDir/package.json`, it creates a `.desktop`. Otherwise, it tells the user (via the console) that `.desktop` file will not be generated.
- Should the file be a requirement?
- Should the inputs configured in linuxCfg be validated?
- Then there's the [`Exec` key](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html) - it should probably be left up to the user?

sysrage (approved):
Winston seems like overkill for this project, but it will be useful to be able to set the `LOG_LEVEL` environment variable to suppress some of the unnecessary messages.

I only scanned through the code, but thank you for removing usage of `fs.*Sync`! 😉 

I didn't have any time to test the `.desktop` file in Linux, but some comments and answers to your questions:

- The `linuxCfg` configuration object should be a property of the `nwbuild` property in `package.json`, not a base property.

- *"Should the `.desktop` file be a requirement for Linux builds?"*
  - No. If a user selects to build a Linux package, it should include a `.desktop` file by default but users should be able to disable the file's creation with a flag in the `nwbuild` property of `package.json` (possibly even an extra property of `nwbuild.linuxCfg`).

- *"Should the inputs configured in linuxCfg be validated?"*

  - Probably some minimal validation, yes. For instance, you wouldn't want to allow them to override the `Type` property and see below about `Exec`.

- *"Should `Exec` key be left up to the user?"*
  - Yes, users need to be able to define this property because they may want to add command-line arguments or something else. However, it should default to the "built" application executable name. Since you currently don't allow renaming this file, it would default to `nw.exe`, `nw`, etc. There should be an `nwbuild` property that causes the executable to be renamed either to the `name` property in `package.json` or a name specified in `nwbuild`.

- *"I've also added code comments and logging in the files changed by this PR"*

  - Winston seems like overkill to me, but it will be useful to be able to disable messages with an environment variable.

- *"Are there any messages which should be moved to other log levels?"*
  - I don't have time to review all the log messages, but suggest that you use extremely minimal logging at INFO or NOTICE. Nearly everything should be a DEBUG, WARNING, or ERROR messages.

PR #694, opened 2022-11-19
feat: add support for excludes

tonetechnician:
Fixes: 

- Fixes issue where module fails if cache directory not already present

### Changes

- add support for exclusions
- include [globby](https://www.npmjs.com/package/globby) library

### DIFF
--- a/src/nwbuild.js
+++ b/src/nwbuild.js
@@ -26,6 +26,7 @@ import { packager } from "./bld/package.js";
 const nwbuild = async ({
   srcDir,
   cacheDir = "./cache",
+  excludes,
   version,
   flavour,
   platform,
@@ -62,6 +63,10 @@ const nwbuild = async ({
     }
   }
 
+  if (cacheDir) {
+    await fs.mkdir(cacheDir, { recursive: true });
+  }
+
   let releaseInfo = await getReleaseInfo(version, cacheDir, manifestUrl);
   let nwDir = `${cacheDir}/nwjs${
     flavour === "sdk" ? "-sdk" : ""
@@ -85,7 +90,7 @@ const nwbuild = async ({
   if (run === true) {
     await develop(srcDir, nwDir, platform);
   } else {
-    await packager(srcDir, nwDir, outDir, platform, zip, releaseInfo);
+    await packager(srcDir, nwDir, outDir, platform, zip, releaseInfo, excludes);
   }
 };
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,10 +13,10 @@
         "cli-progress": "^3.11.2",
         "extract-zip": "^2.0.1",
         "fast-glob": "^3.2.12",
-        "globby": "^13.1.2",
         "plist": "^3.0.6",
         "progress": "^2.0.3",
         "rcedit": "^3.0.1",
+        "recursive-copy": "^2.0.14",
         "tar": "^6.1.11",
         "yargs": "^17.5.1"
       },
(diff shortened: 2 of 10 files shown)


tonetechnician:
Probably needs a bit of formatting but early version of exclusions which I think is super useful and was something I relied on with nwjs-builder-phoenix

tharatau:
Thanks for this! I'll be afk for most of today, will try reviewing in the evening. What do you think of having a `noGlob` flag which defaults to `true`? Turning off the flag would enable globbing how it was in v3?

tonetechnician:
All good, the current PR needs some more testing probably but have been using it with my own program and it's working.

noGlob sounds great! Then we rather keep the older behaviour of passing globs in an array? 

I can remove my changes but keep the cache fix.

Will only get back to it tomorrow most likely.

src/nwbuild.js
```diff
@@ -62,6 +63,10 @@ const nwbuild = async ({
     }
   }
 
+  if (cacheDir) {
+    await fs.mkdir(cacheDir, { recursive: true });
+  }
+
```
tharatau, line 69: Wow your solution is much more elegant than mine god damn

tonetechnician: Haha, seems straightforward enough.

I think my glob solution needs more work though. Currently needing to pass in the name of the file and it's contents glob pattern to exclude the file... 

Also have some more work in that renames the .app product once it's completed building, but need to test on windows

tharatau: > I think my glob solution needs more work though. Currently needing to pass in the name of the file and it's contents glob pattern to exclude the file...

I was thinking of a way to avoid a glob flag. For example: `./src` could imply `./src/**/*` if they don't have any specific globbing patterns in mind? To prevent a breaking change.we could pass multiple patterns (`./src ./assets/*`) as a string.

> Also have some more work in that renames the .app product once it's completed building, but need to test on windows

I'm actually working on that in #712

tharatau:
@tonetechnician Apologies for the back and forth in my PR comments. The reason I dropped globing in v4 was to focus on getting builds right, adding nw-gyp, nwjc, etc. Since tools like `gulp` and similar exist, I wanted to leave what to add in `srcDir` to the developer.

@TheJaredWilcurt @sysrage Thoughts?

TheJaredWilcurt:
Ultimately, for automated builds we will need the ability to exclude copying specific files to the output folder. Also I would need the ability to define a set of "junk files" that will be removed after an `npm install`. There are a lot of useless files that get downloaded during `npm install` that don't need shipped to the end user, such as `README`'s, `/tests` folders, etc.

When I started down this road, the API I designed looked like this:

* https://github.com/nwutils/nw-utils-builder

My current plan is to wait for `nw-builder` to get to a more stable place, then rename `nw-utils-builder` to `nw-builder-task-runner` and use `nw-builder` as a dependency to take advantage of anything it can do for me, while supplementing out the rest of my already designed API. The more of the features it can handle for me, the better.

sysrage:
Agree with much of Jared's comments. Deciding what files to include in a build is actually one of the biggest reasons I haven't made my own build tool generic/published. Every project seems to need different files in a different structure, making a simple "include X files" insufficient.

Ultimately, it would be ideal to be able to provide an array of sources and destinations matched to the distribution architecture. Even further-expanding that with possible renames. Something like:

```json
"includes": {
  "all": [
    {
      "source": "../../dist-files/package.json.dist",
      "destination": "/package.json"
    },
    {
      "source": "./build/**/*",
      "destination": "/build/"
    }    
  ],
  "windows": [
    {
      "source": "../../dist-files/windows/**/*",
      "destination": "../bin/"
    },
  ],
  "linux": [
    {
      "source": "../../dist-files/linux/**/*",
      "destination": "../bin/"
    },  
  ]
}
```

- `../../dist-files/package.json.dist' would be renamed to `package.json` and put in `app/package.nw` directory.
- All files and directories inside `./build/` would be put in the `app/package.nw/build/` directory.
- **Windows builds only**: All files and directories inside `../../dist-files/windows` would be put in the `app/bin/` directory.
- **Linux builds only**: All files and directories inside `../../dist-files/linux` would be put in the `app/bin/` directory.

sysrage:
Regarding Jared’s comments about included `node_modules`:

- The build process should optionally run `npm ci` in the build directory to install the specific version of dependencies in `package-lock.json`. This obviously requires copying `package-lock.json` then deleting it again when packages have been installed.
- Users should just be able to specify an optional script to run that will “finalize” the build. This would run before the “zip” or “package” steps and allow cleanup steps like Jared mentioned. Maybe use ENV variables to let the script know where to find files.

TheJaredWilcurt:
I don't have a need for that level of granularity (array of input/output), but do like the design, as it future proofs it and handles my simpler case, and allows for more advanced cases. And maybe with the ability to copy/rename made so easy I would start to take advantage of it.

I'm weary of environment variables. They tend to be used as bandaids when better solutions could be done and have a lot of cross-platform issues. Though `cross-env` node module tends to work pretty well at solving the cross-platform problems that come with ENV.

Similarly I've had issues running `npm ci` on Windows (very project dependent). Though, if it worked, that would be the preferred approach (even with the copying/deleting of the package-lock).

Hooks for custom scripts to run is a good idea.

tharatau:
@tonetechnician I've added file globing which _should_ take care of your use case.

@TheJaredWilcurt File globing is the first step in allow listing which file and directories to include in the `outDir`.

> Ultimately, it would be ideal to be able to provide an array of sources and destinations matched to the distribution architecture. 

@sysrage As Jared said, I also don't think there is a need for that level of granularity. Each `nwbuild` function builds for one platform-arch. I believe one can implement your use case using file globing.

Thanks for your comments and apologies for getting back to you so late.

PR #712, opened 2022-11-26
fix: rename nw exe

tharatau:
### Changes

- Rename NW executable to `name` property of `srcDir/package.json`


tharatau:
@sysrage in [v3.8.6](https://github.com/nwutils/nw-builder/blob/f647de098ddfc0ef400ea8041be2cedc4be3622a/lib/index.cjs#L489) the NW executable is renamed to the `appName` argument. What do you think of changing it to the value in the `package.json`'s name property?

sysrage:
> @sysrage in [v3.8.6](https://github.com/nwutils/nw-builder/blob/f647de098ddfc0ef400ea8041be2cedc4be3622a/lib/index.cjs#L489) the NW executable is renamed to the `appName` argument. What do you think of changing it to the value in the `package.json`'s name property?

In my opinion, this should default to `name` from `package.json`. However, users should be able to override that name with `nwbuild.appName` (or something similar).

src/bld/package.js
```diff
@@ -43,6 +44,9 @@ const packager = async (srcDir, nwDir, outDir, platform, zip, releaseInfo) => {
   log.debug("Convert package.json buffer into JSON");
   let pkg = JSON.parse(buffer);
 
+  log.debug(`Rename NW executable to ${pkg.name}`);
+  await rename(`${outDir}/${getPlatformSpecificName(platform)}`, `${outDir}/${pkg.name}`);
```
sysrage, line 48: Isn't this going to incorrectly remove the `.exe` extension for Windows and `.app` extension for macOS?

sysrage (changes requested):

src/bld/package.js
tharatau: Should be fixed now

sysrage: I don't see any changes made to fix this? Still renaming to `${outDir}/${pkg.name}`.

tharatau: It should rename to `${outDir}/${pkg.name}` for Linux, `${outDir}/${pkg.name}.app` for Mac and `${outDir}/${pkg.name}.exe` for Windows

sysrage: My apologies. Not sure what I was looking at last night, but I see the correct code now.

src/bld/winCfg.js
```diff
@@ -1,7 +1,18 @@
...
+ *
+ * @param {object} pkg     The srcDir/package.json as a JSON
+ * @param {string} outDir  The directory to hold build artifacts
+ */
 const setWinConfig = async (pkg, outDir) => {
-  return rcedit(`${outDir}/nw.exe`, {
+
+  await rename(`${outDir}/nw.exe`, `${outDir}/${pkg.name}.exe`);
+
+  rcedit(`${outDir}/${pkg.name}.exe`, {
```
sysrage, line 15: `rcedit` returns a promise. You need to `await` this line.

sysrage (changes requested):
I still feel users should be able to specify a name other than the `name` property in `package.json`. Users should be able to add an `appName` (or some other property) to the `nwbuild` configuration object, which would override the default `name`.

tharatau:
> I still feel users should be able to specify a name other than the `name` property in `package.json`. Users should be able to add an `appName` (or some other property) to the `nwbuild` configuration object, which would override the default `name`.

I agree that there should be a way to change app name, but I'd like to keep that for another PR. I first want to figure out the platform specific cfgs and after they all work, let user customise them.

sysrage (approved):

PR #718, opened 2022-11-28
Dev 704

tharatau:
Fixes: #

### Changes

- ...


docs/index.md
```diff
@@ -97,13 +97,13 @@ package.json usage
...
-| flavour     | `sdk      | normal`                    |                                                                                | NW runtime build flavour.                                            |
-| platform    | `linux    | osx                        | win`                                                                           |                                                                      | NW supported platforms |
-| arch        | `ia32     | x64`                       |                                                                                | NW supported architectures                                           |
+| ----------- | --------- | -------------------------- | ------------------------------------------------------------------------------ | -------------------------- | ---------------------- |
+| srcDir      | `string`  |                            | Directory to hold NW app files unless or array of file glob patterns           |
```
sysrage, line 101: - The `srcDir` description is a little confusing and the type is wrong if it allows either a string or array of strings.
- Several missing descriptions below.
- The formatting of this table is also busted and doesn't display properly on GitHub.

src/get/getReleaseInfo.js
```diff
@@ -28,6 +28,10 @@ export const getReleaseInfo = async (version, cacheDir, manifestUrl) => {
     log.debug("Convert manifest data into JSON");
     let manifestJson = JSON.parse(manifestData);
     log.debug(`Search for ${version} specific release data`);
+    if (version === "latest" || version === "lts") {
+      version = manifestData[version].slice(1);
```
sysrage, line 32: Reassigning function parameters! 😢 

I don't think "lts" makes sense for anything NW.js-related. Unfortunately, there aren't any LTS builds available and they don't even match the Node.js LTS builds. So, supporting an "lts" version is misleading.

I also noticed that you mentioned a "stable" version being supported in the PR comments. This would be the same as "lts", where it simply doesn't exist.

src/nwbuild.js
```diff
@@ -1,96 +1,148 @@
...
+      );
+      await decompress(options.platform, options.cacheDir);
+      await remove(options.platform, options.cacheDir);
+    }
+
+    if (options.mode !== "run" && options.mode !== "build") {
+      throw new Error("Invalid mode value. Expected run or build.");
+    }
+    if (options.mode === "run") {
+      await develop(options.srcDir, nwDir, options.platform);
```
sysrage, line 132: Since the only options are "run" or "build", you should probably `return` after this `develop()` call. No need to run through the rest of the function.

As an additional change, you could even remove the `if (options.mode === "build")` (after adding the `return` above). This would make "build" the default mode, so you could also remove the check on lines 128-130.

src/util/arch.js
```diff
@@ -0,0 +1,19 @@
...
+  switch (arch) {
+    case "ia32":
+      return "ia32";
+    case "x64":
+      return "x64";
+    default:
+      log.error(`The arch ${arch} is not supported`);
+      return undefined;
+  }
+};
```
sysrage, line 19: No need for this function to be `async` and using a `switch()` just feels wrong...

```js
export const getArch = (arch) => {
  if (!['ia32', 'x64'].includes(arch)) {
    log.error(`The arch ${arch} is not supported`);
    return undefined; // should this be a thrown error instead?
  }
  return arch;
};
```

src/util/arch.test.js
```diff
@@ -0,0 +1,13 @@
...
+    expect(getArch("arm64")).rejects.toThrow();
+});
+
+test("x32 arch support", () => {
+    expect(getArch("ia32")).resolves.toBe("ia32");
+});
+
+test("x64 arch support", () => {
+    expect(getArch("x64")).resolves.toBe("x64");
+});
```
sysrage, line 13: I don't know Jest but `getArch()` shouldn't be async. See my comments in `arch.js`.

src/util/platform.js
```diff
@@ -0,0 +1,21 @@
...
+  switch (platform) {
+    case "darwin":
+      return "osx";
+    case "win32":
+      return "win";
+    case "linux":
+      return "linux";
+    default:
+      log.error(`The platform ${platform} is not supported`);
+      return undefined;
```
sysrage, line 19: Throw an error? This also shouldn't be `async`.

src/util/validate.js
```diff
@@ -0,0 +1,11 @@
...
+ * Validate options
+ *
+ * @param  {object}  options      Options
+ * @param  {object}  releaseInfo  Version specific NW release info
+ * @return {boolean}              True if options are valid. False otherwise
+ */
+export const validate = async (options, releaseInfo) => {
+  if (options && releaseInfo) return true;
+  else return false;
+};
```
sysrage, line 11: This also shouldn't be `async`. Lots of these that need to be fixed.

sysrage (changes requested):

src/get/getReleaseInfo.js
sysrage: Another thought, can `latest` become the default if users didn't specify a version?

docs/index.md
tharatau: Fixed the type and description for `srcDir`. Also formatted the table correctly

src/get/getReleaseInfo.js
tharatau: LTS in this case refers to NW.js v0.14.7. I believe this is the last version to support Windows XP. The current NW release flow matches the latest Node version with the latest Chromium version.

p.s.: I am planning to [do custom builds](https://github.com/tharatau/nw/issues/1) but setting that up will take some time.

src/util/arch.js
tharatau: ✅

src/util/arch.test.js
tharatau: ✅

src/util/validate.js
tharatau: ✅

src/util/platform.js
tharatau: ✅

src/get/getReleaseInfo.js
sysrage: I don't agree with calling 0.14.7 "LTS", but that's your call. That version is simply not "long term support". It's not supported anymore and it doesn't work like any remotely-modern version. If anything, "legacy" would be a more-fitting name.

Either way, what are you considering "stable"? There is no branch that's considered "stable", there's only "latest".

src/nwbuild.js
sysrage: So, I haven't had a chance to review the changed code yet, but I see the new documentation still doesn't have defaults for `mode`, `version`, or `flavour` (note that you should probably support either `flavor` or `flavour`). I think all of these properties should have defaults:
`mode`: default to "build"
`version`: default to "latest"
`flavor`: default to "normal"

Can probably even get away with defaults for `platform` (default to either the current OS or every platform), `arch` (default to "x64"), and `outDir` (default to "./dist/").

src/get/getReleaseInfo.js
sysrage: I just looked at `https://nwjs.io/versions` and see where you're getting the lts/stable from. Those really don't make sense for how Roger releases NW.js, but use whatever you want!

tharatau: > I don't agree with calling 0.14.7 "LTS"

I agree but I think that is an upstream issue.

> If anything, "legacy" would be a more-fitting name.

I'll leave lts and stable in for now. I'll open an issue upstream.

src/nwbuild.js
tharatau: Is it best practice for API properties to have defaults?

The reason I didn'twant these arguments to have default values is because the "default" is derived from the `https://nwjs.io/versions` manifest file. A feature of nw-builder is the ability to define custom manifest files with their own defaults. That's why I thought it would not makes sense to have default values for only one type of manifest.

sysrage (approved):
It's worth having further discussion on the defaults, but everything else looks fine.

src/nwbuild.js
sysrage: > **NW.js Arch default:** 32-bit on Windows, 64-bit on Linux/OSX

What?? 64-bit everywhere...

tharatau: > Yes it is better to have defaults for everything, so as the API grows, your usage of it only requires overriding the parts you need to customize. Also as users upgrade the library to newer versions, presumably new API options will be added, so they shouldn't have to update all their code every time to add in default values for new properties.

This makes so much sense!

> **NW.js Arch default"**: 32-bit on Windows, 64-bit on Linux/OSX

Currently it gets the default architecture from `process.arch` so I don't think hardcoding a default arch is necessary.

tharatau:
> It's worth having further discussion on the defaults, but everything else looks fine.

I've opened another issue:
https://github.com/nwutils/nw-builder/issues/727

src/nwbuild.js
tharatau: Good to know!

Issue #727, opened 2022-12-10
Set defaults for options


tharatau:
### Issue Type

- [ ] Bug Fix
- [x] Feature
- [ ] Other

### Additional Info

- Operating System:
- Node version:
- NW.js version:
- Repro link:
- ...

Issue #706, opened 2022-11-25
Incorrect configuration option for Windows EXE icon


sysrage:
### Issue Type

- [X] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour

https://github.com/nwutils/nw-builder/blob/3090addc3931c3ab5edea0e5afccfe613ed63b24/src/bld/winCfg.js#L7

The icon being used for `rcedit` appears to be obtained from the base "icon" property in `package.json`.

- There is no base "icon" property in `package.json`.
- If this is meant to be `window.icon`, the two icon formats do not match (`rcedit` needs an ICO while `window.icon` needs a PNG).

### Expected/Proposed Behaviour

There should be a separate property of the `nwbuild` configuration object, which specifies the icon that should be used with `rcedit`. Something along the lines of:
```js
{
  "nwbuild": {
    "winIcon": "./appicon.ico"
  }
}
```

Issue #699, opened 2022-11-20
Linux builds fail if `nwbuild.linuxCfg` is not set


sysrage:
### Issue Type

- [X] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour

Optional chaining isn't being used, which causes a failure if `nwbuild.linuxCfg` is not defined:
```
file:///C:/Users/Brian/Documents/Code/nw-react/node_modules/nw-builder/src/bld/linuxCfg.js:20
  if (typeof pkg.nwbuild.linuxCfg === "object") {
                         ^
```

### Expected/Proposed Behaviour

Use optional chaining:
```js
if (typeof pkg.nwbuild?.linuxCfg === "object") {
```

tharatau:
Wow I literally remember doing that, must have made a typo

sysrage:
So, it's actually even worse than just the missing optional chaining. This fails if you don't have `nwbuild` defined in `package.json` at all. I'm running this by passing all configuration objects as a parameter, not via CLI. If you don't have an `nwbuild` property in `package.json`, you get the error above. If you have `nwbuild` (even if it's an empty object), you still get the error above unless you have a `linuxCfg` prop or add the optional chaining fix.

Please fix this so you don't need `nwbuild` in `package.json` at all. The `linuxCfg` property should be optional, whether you're using `package.json` or passing a configuration object as a parameter to `nwbuild()`.

tharatau:
> Edit: Second half of this is that you need to allow specifying linuxCfg as part of the configuration object passed to nwbuild(). As a workaround to the above issue, I tried adding this property but it didn't work.

`linuxCfg` as a nwbuild option definitely makes more sense.

Issue #695, opened 2022-11-20
Add configuration property to rename NW.js executable


sysrage:
### Issue Type

- [ ] Bug Fix
- [X] Feature
- [ ] Other

### Current/Missing Behaviour

Please add a property to the `nwbuild` configuration object which allows users to specify a name that `nw.exe` (and the matching binary for all other operating systems) should be renamed as.

tharatau:
I'm tackling this platform by platform, next PR will be for Windows config using `rcedit`

sysrage:
Unlinked the PR, so this remains open. The PR is a good stop-gap, renaming the the name specified as `name` in `package.json`. Users still need a way to choose what name to use, though.

PR #729, opened 2022-12-13
[wip] feat: linux and windows cfg options

tharatau:
Fixes: #695
Fixes: #699
Fixes: #706
Fixes: #727

### Changes

- Set defaults for remaining options
- Rename nw executable to `app.name` and fallback to `package.json`'s `name`
- Add Linux desktop entry file properties as options
- Add Windows config options to `rcedit`


src/util/parse.js
```diff
@@ -23,15 +19,54 @@ export const parse = async (options) => {
...
-  options.srcDir = options.srcDir ?? undefined;
-  options.version = options.version ?? undefined;
-  options.flavour = options.flavour ?? undefined;
-  options.platform = options.platform ?? undefined;
-  options.arch = options.arch ?? undefined;
-  options.outDir = options.outDir ?? undefined;
+  options.srcDir = options.srcDir ?? "./";
+  options.mode = options.mode ?? "build";
+  options.version = options.version ?? "latest";
+  options.flavour = options.flavour ?? "sdk";
```
TheJaredWilcurt, line 25: `'normal'` would likely be the desired default flavor for a build in almost all circumstances.

Cases I can think of where I'd want SDK:

1. If I did a "Stable" and "Beta" or "Canary" release of an app, the beta/canary version may ship with dev tools enabled, while the "Stable" wouldn't.
1. I might do an sdk build locally manually to debug, check, or validate something with the build, though this would be fairly rare.
1. Some apps are specifically aimed at developers, (like Scout-App), and want to encourage tinkering with the app itself.

I can't really think of any other scenario. Most NW.js app creators just ship the 'normal' version.

```suggestion
  options.flavour = options.flavour ?? "normal";
```
If this default is changed, the documentation table in `docs/index.md` would need updated as well.

test/demo/nwapp/package.json
```diff
@@ -1,4 +1,6 @@
 {
     "name": "nwdemo",
-    "main": "index.html"
+    "version": "0.0.1",
+    "main": "index.html",
+    "author": "ABC Corp"
 }
```
TheJaredWilcurt, line 6:
```suggestion
}

```

src/util/parse.js
TheJaredWilcurt: Also, most English speaking countries follow the America variations for `color` and `flavor`, as opposed to the British spelling. I'm actually fine if the API supports both. But specifically all documentation and API for NW.js uses the American style.

* https://nwjs.io/versions.json
* `console.log(nw.process.versions['nw-flavor'])`
* https://nwjs.readthedocs.io/en/latest/For%20Users/Advanced/Build%20Flavors/#build-flavors

```suggestion
  options.flavour = options.flavour || options.flavor || "normal";
```

Also changed to OR, since the flavor can only be normal or sdk, so no reason to allow passing in other falsy values with nullish co.

src/bld/linuxCfg.js
```diff
@@ -6,30 +6,48 @@ import { log } from "../log.js";
...
+  await rename(`${outDir}/nw`, `${outDir}/${app.name}`);
+
   let fileContent = `[Desktop Entry]\n`;
   Object.keys(desktopEntryFile).forEach((key) => {
-    fileContent += `${key}=${desktopEntryFile[key]}\n`;
-    log.debug(`Add ${key}=${desktopEntryFile[key]} to Desktop Entry File`);
+    if (desktopEntryFile[key] !== undefined) {
+      fileContent += `${key}=${desktopEntryFile[key]}\n`;
+      log.debug(`Add ${key}=${desktopEntryFile[key]} to Desktop Entry File`);
+    }
```
TheJaredWilcurt, line 48: This PR make a lot of changes to this file. Most of the file's code is focused on creating a desktop shortcut. The NW-Utils GitHub organization already has a library that handles that:

* https://github.com/nwutils/create-desktop-shortcuts

It additionally sets the chmod on the shortcut to allow the shortcut to be executed. Since that library is already fully unit tested, we should pull it in here to reduce the amount of code in this file and shift the burden of testing this logic over to a codebase scoped to handle just that feature.

Also there are several parameters listed in this PR that are not recognized in all Linux distros. So they are not included in the `create-desktop-shortcut` API. I'm not sure if any would actually be relevant to NW-Builder users. But if so we should probably add them to the `create-desktop-shortcuts` library anyway.

src/util/parse.js
tharatau: Agree with `normal` as default flavour. Leaning towards keeping it American English. I spelt it as `flavour` out of habit but keeping it consistent makes more sense.

src/bld/linuxCfg.js
tharatau: Good to know! I'll have a look at it soon.

TheJaredWilcurt: Can be handled in a future PR as well, I don't want to slow progress on this project.

src/nwbuild.js
```diff
@@ -72,10 +72,18 @@ export const nwbuild = async (options) => {
     }
 
     // Parse options, set required values to undefined and flags with default values unless specified by user
-    options = await parse(options);
+    options = await parse(options, nwPkg);
+
+    // Variable to store nwDir file path
+    nwDir = `${options.cacheDir}/nwjs${
+      options.flavour === "sdk" ? "-sdk" : ""
+    }-v${options.version}-${options.platform}-${options.arch}`;
```
TheJaredWilcurt, line 80: This feels like it's trying to do too much in a compact space. I tried breaking it up on multiple lines. Feel free to ignore/resolve this comment.
```suggestion
    nwDir = [
      options.cacheDir,
      "/nwjs",
      options.flavour === "sdk" ? "-sdk" : "",
      "-v" + options.version,
      "-" + options.platform,
      "-" + options.arch
    ].join('');
```

src/util/parse.js
```diff
@@ -23,15 +19,54 @@ export const parse = async (options) => {
     options.flavour = options.flavor;
   }
```
TheJaredWilcurt, line 20: I didn't notice lines 18-20 before. But they are now redundant due to line 25 and can be removed.

docs/index.md
```diff
@@ -102,13 +102,13 @@ package.json usage
...
-| platform    | `linux \| osx \| win`               |                            | NW supported platforms                                                         |
-| arch        | `ia32 \| x64`                       |                            | NW supported architectures                                                     |
-| outDir      | `string`                            |                            | Directory to store build artifacts                                             |
+| srcDir      | `string`                            | `./`                       | Directory to hold NW app files                                                 |
+| mode        | `run \| build`                      | `build`                    | Run or build application                                                       |
+| version     | `latest \| stable \| string \| lts` | `latest`                   | NW runtime version                                                             |
+| flavour     | `sdk \| normal`                     | `sdk`                      | NW runtime build flavour.                                                      |
```
TheJaredWilcurt, line 108: Since the default changed in the code, it should be reflected in the documentation
```suggestion
| flavour     | `sdk \| normal`                     | `normal`                   | NW runtime build flavour.                                                      |
```

package.json
```diff
@@ -31,9 +31,10 @@
     "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
     "docs": "jsdoc ./src/nwbuild.js -d docs",
     "demo:cli": "cd ./test/demo && nwbuild ./nwapp --mode=build --version=0.70.1 --flavour=normal --platform=win --arch=x64 --outDir=./build",
-    "demo:bld": "cd ./test/demo && node bld.cjs",
-    "demo:run": "cd ./test/demo && node run.cjs",
-    "cjs": "esbuild ./src/nwbuild.js --bundle --minify --legal-comments=inline --platform=node --outfile=./nwbuild.cjs",
+    "demo:nux": "npm run cjs && cd ./test/demo && node nux.js",
```
sysrage, line 34: IMO "nux" is a terrible name for Linux/UNIX (often abbreviated as "*nix" but never "*nux"). I recommend sticking with "linux".

src/bld/winCfg.js
```diff
@@ -4,24 +4,43 @@ import rcedit from "rcedit";
...
+    }
+   });
+
+  await rename(`${outDir}/nw.exe`, `${outDir}/${app.name}.exe`);
+  await rcedit(`${outDir}/${app.name}.exe`, {
+    "file-version": app.version,
+    "icon": app.icon,
+    "product-version": app.version,
+    "version-string": versionString,
+    "requested-execution-level": "asInvoker",
```
sysrage, line 43: Why are you setting `requested-execution-level` here? I don't think this is a good default, as it'll cause a UAC prompt for any admin user. Most apps do not need admin rights. That said, users should be able to specify if their app needs admin access via configuration properties.

src/nwbuild.js
sysrage:
```js
    const flavorString = options.flavour === "sdk" ? "-sdk" : "";
    nwDir = `${options.cacheDir}/nwjs${flavorString}-v${options.version}-${options.platform}-${options.arch}`;
```
IMO, this is more readable than Jared's solution. 

I also notice you aren't using `node:path` everywhere in this project and really should be.

src/nwbuild.js
```diff
@@ -93,27 +101,17 @@ export const nwbuild = async (options) => {
...
     if (options.mode === "run") {
-      let tmpPlatform = getPlatform(platform);
-      let tmpArch = getArch(arch);
-      if (tmpPlatform === undefined) {
+      if (options.platform === undefined) {
         throw new Error(`Platform ${platform} is not supported. Sorry!`);
       }
-      if (tmpArch === undefined) {
+      if (options.arch === undefined) {
         throw new Error(`Architecture ${arch} is not supported. Sorry!`);
```
sysrage, line 108: Why would an error be thrown in "run" mode if either `platform` or `arch` is undefined? Theoretically, because of changes to `parse()`, these would never actually be undefined. However, if a user is attempting to run the app, they would be trying to use the currently running OS. So, shouldn't `platform` and `arch` **always** match the current OS and architecture in "run" mode, even if something different is specified in `package.json` or the `nw-builder` configuration object?

sysrage (changes requested):

src/nwbuild.js
TheJaredWilcurt: Platform, yes. But on Windows 64, it is more common to run 32-bit software than 64-bit. So they may intentionally want to run a 32-bit version of NW.js to ensure that all dependencies work correctly.

BrianBothwell: > Platform, yes. But on Windows 64, it is more common to run 32-bit software than 64-bit.

This is simply not true and I don't understand why you keep trying to propagate it. It is days away from the year 2023. It is absolutely not more common to be running 32-bit and should **not** be done in most cases. Why would you limit your apps memory and other functionality by default, when 64-bit has been the standard for nearly a decade?

TheJaredWilcurt: Though very uncommon to be running a 32-bit *OS*, most Windows software that is installed today is still 32-bit. This is pretty easy to check just by comparing the amount of items in `C:\Program Files` versus `C:\Program Files (x86)`. It's of course going to depend on the user, but across my machines about 65% of the software is still 32-bit.

> Why would you limit your apps memory and other functionality by default?

None of my apps require much resources. If I was making software that taxed the hardware, I'd offer a 64-bit version of my app for Windows. Or if my app relies on a dependency that only supports 64-bit, then I'd be locked out of supporting 32-bit. But otherwise, if my app does not benefit from 64-bit, then I use 32 by default. It is functionally no different to me or my users and will work on more systems and I have one less build to do. Sure there aren't many 32-bit Windows users out there, but realistically there are likely more rare 32-bit Windows users than all non-windows users combined. I use NW.js to get wider platform support for free. If there's no downside to how my app runs, then I'll publish it in the most universal way possible to support the most platforms.

You should really be asking yourself, what benefit do you get by excluding this subset of users. If it requires additional effort, then by all means, skip them, but for most cases, it's the same level of effort to support 32 & 64 bit Windows as it is to only support 64-bit. So might as well do both. For example, I made an app a year or two ago, and it required me to build a native dependency. I decided that until someone requested a 32-bit build, that I wouldn't put forth the effort to support both, as no one would likely request it. I instead put that effort towards finding another native dependency that worked on OSX, building, testing, and releasing a version for that OS (that then no one downloaded). But as a side effect it worked for Linux too, and nearly half of my downloads are for Linux now. So I guess it worked out in the end.

src/bld/linuxCfg.js
tharatau: I have created a separate issue for it.

src/nwbuild.js
tharatau: I think this works for now as long as the manifest format is not changed.

src/bld/winCfg.js
tharatau: Added it in since I thought it was a required property. Just reread the docs for this realising its optional.

src/nwbuild.js
tharatau: > Why would an error be thrown in "run" mode if either platform or arch is undefined?

Before adding in error handling, `undefined` conveyed that the platform/arch is not supported. For example, if a Linux arm64 user downloads Linux amd64 binary, trying to run it would fail with an unsupported arch message.

Currently, it looks like these functions have become redundant. Instead, the validate function should check for platform-arch support.

sysrage: My two cents:
- For "run" mode, use the current OS for `platform` no matter what's in `package.json`. For `arch`, you both make a good point about needing to fail if a user's running Linux arm64. Might want to consider automatically failing over from arm64 to x64 for osx, until a native build is available.
- For any "build" modes, I'm clearly out-numbered but still think defaults should be x64 unless otherwise specified. It'll never make sense to me to sacrifice performance/features to support a tiny fragment of potential users. If 32-bit support is actually needed by your users, by all means support it. Otherwise, it's time to move forward. Let's stop contributing to the `Program Files (x86)` content...

sysrage (approved):

Issue #714, opened 2022-11-27
Building `windows` package in Linux or macOS fails without WINE


sysrage:
### Issue Type

- [X] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour

If you attempt to build a `windows` package in either Linux or macOS, the build fails if WINE is not installed.

### Expected/Proposed Behaviour

It's expected for `rcedit` to not work in these operating systems without WINE. However, `nw-builder` needs to fail gracefully in this case. It should show a warning that the EXE could not be modified, but continue on with the build.

Additionally, the WINE requirement should be documented.

PR #739, opened 2022-12-25
fix(win-bld): skip modify exe if platform is not windows or wine is not installed

tharatau:
Fixes: #714

### Changes

- handle `rcedit` error on missing wine or if platform is not windows

### DIFF
--- a/src/bld/winCfg.js
+++ b/src/bld/winCfg.js
@@ -1,5 +1,4 @@
 import { rename } from "node:fs/promises";
-import { platform } from "node:process";
 
 import rcedit from "rcedit";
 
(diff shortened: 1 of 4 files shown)


src/bld/winCfg.js
```diff
@@ -34,13 +37,24 @@ const setWinConfig = async (app, outDir) => {
...
-  await rcedit(`${outDir}/${app.name}.exe`, {
-    "file-version": app.version,
-    "icon": app.icon,
-    "product-version": app.version,
-    "version-string": versionString,
-  });
+  try {
+    await rename(`${outDir}/nw.exe`, `${outDir}/${app.name}.exe`);
+
+    if (platform === "win32") {
```
sysrage, line 43: Isn't `process.platform` going to not equal "win32" when running in macOS or Linux with WINE? I don't think this check will work in those operating systems.

I think you can probably get away with something like:

```js
try {
    await rcedit(...);
} catch (error) {
    throw new Error(`Unable to modify EXE. Ensure WINE is installed or build in Windows.\nError: ${error}`);
}
```

sysrage (changes requested):

src/bld/winCfg.js
tharatau: I didn't `throw` the error since the error is not caught in the parent function.

sysrage (approved):

PR #749, opened 2022-12-30
fix: add back globing

tharatau:
### Description

The `nwbuild` function parses the globing patterns in `srcDir` (for eg: `srcDir: "./src/**/* .dist/**/*"`) and saves the list of file names in the `files` variable. The `packager` function uses the array to copy the relevant files to the `outDir`.


src/nwbuild.js
```diff
@@ -77,41 +80,47 @@ import { log } from "./log.js";
...
+    let files = [];
+    let patterns = options.srcDir.split(" ");
+
+    for (const pattern of patterns) {
+      let contents = await glob(pattern);
+      for (const content of contents) {
+        if (basename(content) === "package.json" && nwPkg === undefined) {
+          nwPkg = JSON.parse(await readFile(content));
+        }
+        files.push(...contents);
```
davedoesdev, line 97: Shouldn't this go outside the loop?
I'm getting 

```
[ ERROR ] 2023-01-05T08:19:05.751Z Invalid array length
```

with it inside the loop when I have a large number of matched files.

sysrage: Additionally, this seems to conflict:

You describe `srcDir` as: "Directory to hold NW app files unless or array of file glob patterns"

But then you do this:
https://github.com/nwutils/nw-builder/blob/af32f1095578ac35c2cabcc7186ef8b4986277cb/src/nwbuild.js#L89

If a user provides an array of glob patterns (like you document is allowed), it will fail.

Issue #766, opened 2023-01-12
Notify function resolving wrong `package.json` path


dominickbrasileiro:
### Issue Type

- [X] Bug Fix
- [ ] Feature
- [ ] Other

### Current/Missing Behaviour
The nw-builder CLI is resolving the wrong package.json path when trying to notify if the package has available updates.

In `./src/util/notify.js`, `packageJson` is resolving to my home directory because it's relative to the path that I run the CLI.:
```js
const packageJson = JSON.parse(
  await readFile(resolve("..", "..", "package.json")),
);
```

Error:
![image](https://user-images.githubusercontent.com/50509793/212168990-977fe6b4-7234-4e35-801c-2d380e596838.png)

### Proposed Solution
Instead of resolving the package.json's path, we can import it directly:
```js
import updateNotifier from "update-notifier";

import packageJson from "../../package.json" assert { type: "json" };

/**
 * Notify the user if there is a new version of the package available.
 *
 * @return {Promise<void>}
 */
export const notify = async () => {
  updateNotifier({ pkg: packageJson }).notify();
};
```

### Additional Info

- Operating System: Ubuntu 22.04
- Node version: v18.13.0
- nw-builder version: 4.0.6

PR #767, opened 2023-01-12
fix: correct notify util

dominickbrasileiro:
Fixes: #766

### Changes

- import the packageJson directly instead of reading it in runtime.

### DIFF
--- a/src/util/notify.js
+++ b/src/util/notify.js
@@ -1,12 +1,15 @@
+import { createRequire } from 'node:module';
+const require = createRequire(import.meta.url);
+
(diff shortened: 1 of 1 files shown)


tharatau:
Thanks for this! Once the lint error is fixed (disabled?), this is good to go.

dominickbrasileiro:
Unfortunately, the linter fails because import assertions were introduced in node v17.1.0 according to [this](https://nodejs.org/api/errors.html#err_import_assertion_type_missing). I had to work around it using the `require` and `createRequire` functions, so node >= 12 will be supported.

PR #785, opened 2023-02-04
fix: file globing

tharatau:
Fixes: #

- [x] Format
- [x] Lint
- [x] Test
- [x] Docs
- [ ] Add name to AUTHORS (first time contributors)
- [x] Update CHANGELOG
- [x] Version bump

### Description

Undocumented bug discovered through [Stack Overflow](https://stackoverflow.com/questions/75275953/nw-js-node-webkit-error-package-json-not-found-in-srcdir-file-glob-patterns/75341913#75341913). `node-glob` and yargs file glob seem to [work differently](https://github.com/yargs/yargs/issues/2294). For now, one can tweak the glob pattern for CLI to match the required files and directories. An `options.cli` flag is set to `true` on each CLI run. This prevents `node-glob` from doing twice the work and erroring out with the error message as mentioned in the Stack Overflow post.

With module usage, the copied file `subdir/file.js` would be copied as `outDir/package.nw/file.js`. This is now fixed and is copied as `outDir/package.nw/subdir/file.js`.


src/bld/package.js
```diff
@@ -34,13 +34,13 @@ const packager = async (
...
 
-  for (const file of files) {
+  for (let file of files) {
     log.debug(`Copy ${file} file to ${outDir} directory`);
     await cp(
       file,
       `${outDir}/${
         platform !== "osx" ? "package.nw" : "nwjs.app/Contents/Resources/app.nw"
-      }/${basename(file)}`,
+      }/${file.split(sep).splice(2).join(sep)}`,
```
rickymohk, line 43: This line cause problem on windows. "/" instead of `sep` should be used here (The rest of the path is already using "/". `file` come from glob expanded `srcDir`, which always use "/". `split(sep)` split nothing on windows causing the subdirectory disappear.

tharatau: Can you open a bug report referencing this line with a minimal repro?

tharatau: I'm working on getting Windows builds working in #795. I realise I need to use `path.resolve` which would remove the need for `sep`.

PR #829, opened 2023-03-26
feat: add `glob` flag

tharatau:
Fixes: #

## Description

- Add `glob` flag
- Recommend disabling globbing in docs
- Remove strict runtime check for Node version


sysrage:
Two issues:

The version check fails (not sure why the message is repeated twice, either):
```
[ ERROR ] 2023-03-27T00:00:14.157Z NW.js v0.72.0 requires Node.js v19.3.0 but you are using Node.js v19.8.1.
[ ERROR ] 2023-03-27T00:00:14.157Z NW.js v0.72.0 requires Node.js v19.3.0 but you are using Node.js v19.8.1.
```

When using `glob: false`, something is `undefined` when attempting to copy `srcDir` to `outDir`:

image

sysrage:
All of the changes I needed to make, to get my app to build:

1. [Disable the Node.js version check](https://github.com/nwutils/nw-builder/blob/feat/glob-flag/src/util/validate.js#L13-L17). Even with the right version installed, it was failing.

2. [Disable all of the changes being made to `CFBundleDisplayName`](https://github.com/nwutils/nw-builder/blob/feat/glob-flag/src/bld/osxCfg.js#L26-L48). With these enabled, the resulting `undefined.app` won't run.

3. [Disable automatic writing of `package.json`](https://github.com/nwutils/nw-builder/blob/feat/glob-flag/src/bld/build.js#L72-L83). I have already created an appropriate `package.json`, I don't want `nw-builder` to modify it.

This results in a file named `undefined.app` that will run. It appears that the name is incorrect, because [`pkg.name` is undefined in my use-case](https://github.com/nwutils/nw-builder/blob/503ca5f7f51758b11ce13aae31bc55f16664457a/src/bld/osxCfg.js#L23).

src/util/validate.js
```diff
@@ -10,50 +10,71 @@ import { readdir } from "node:fs/promises";
...
-    }
-    if (releaseInfo.components.node !== version.slice(1)) {
-      return new Error(
-        `NW.js ${releaseInfo.version} requires Node.js v${
-          releaseInfo.components.node
-        } but you are using Node.js ${version.slice(1)}.`,
-      );
-    }
-  } else {
+  if (releaseInfo.components.node > version.slice(1)) {
```
sysrage, line 13: I still recommend removing this check altogether. Most users really don't need it. If you keep it, change it to a warning instead of a critical error please.

sysrage (changes requested):
With these new changes, everything builds correctly as long as you have a new Node.js version installed.

image

src/util/validate.js
tharatau: After giving it some thought I can see why warning is better. I was thinking about Native Node modules but there are a lot of apps which don't use it (also there's isn't a good way to rebuild native node modules).

PR #830, opened 2023-03-27
Fixes missing `files` when calling `getManifest()` with glob disabled

sysrage:
When `glob` is set to `false`, the call to `getManifest()` was failing in Windows:

```
[ ERROR ] 2023-03-27T12:32:48.495Z ENOENT: no such file or directory, open 'C:\package.json'
```

This PR sets `files` to `options.srcDir` if `options.glob` is false. It works when globbing is disabled, but I don't know if it breaks anything when globbing is enabled.

### DIFF
--- a/src/nwbuild.js
+++ b/src/nwbuild.js
@@ -98,7 +98,7 @@ const nwbuild = async (options) => {
     options = await parse(options, manifest);
 
     if (options.mode !== "get") {
-      files = await getFiles(options.srcDir);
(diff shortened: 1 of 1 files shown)


tharatau:
https://github.com/nwutils/nw-builder/blob/5dc7d680137a6d394d9325beb1525d8188dd3bbe/src/nwbuild.js#L232

This doesn't work?

tharatau (approved):
Hmm, I can see why this would happen. Thanks!

Issue #822, opened 2023-03-21
Cache never expires


sysrage:
### Issue Type

- [X] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour

If you have an old `manifest.json` file in your `cache` directory, it's never updated. So, if a new version of NW.js comes out builds will fail.

### Expected/Proposed Behaviour

If using a cache, it needs to expire at some point.

### Additional Info

- Package version: 4.1.1

williamwang96:
would it be an acceptable fix to always download `manifest.json`? It's very lightweight and shouldn't be cached anyways. I can work on a PR if the fix sounds good to you @tharatau .

tharatau:
@williamwang96 My motivation for caching `manifest.json` was to allow builds without network (if the relevant nw binary is already downladed, that is). Ideally we redownload `manifest.json` when connected to network and don't while not connected to network.

PR #888, opened 2023-06-27
fix: change manifest cache impl

ayushmanchhabra:
Fixes: #822

## Description

- Redownload `manifest.json` every time the `nwbuild` function is executed.
- If `manifest.json` already exists and we are unable to connect to the `nwjs.io` domain (network related issues), then use the existing manifest.
- If no `manifest.json` exists in the cache dir, then the validate function will catch this and throw an error - preventing the build.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [4.2.7] - 2023-06-27
+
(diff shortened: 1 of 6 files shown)


ayushmanchhabra:
@sysrage I feel like this is PR strike a good balance w.r.t the discussion in #822.

williamwang96 (approved):
This looks great and is exactly what I was thinking!

PR #891, opened 2023-06-28
docs:  move `doc/license.md` to `/LICENSE` for better visibility

TheJaredWilcurt:
The `package.json` already mentions MIT, this is just being more clear to users of the library.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 NW.js Utilities
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


ayushmanchhabra:
I was actually trying to find the license myself. Don't remember why but I put it inside the `/doc` dir.

ayushmanchhabra (changes requested):
Instead of creating this file, we could move the license from `/doc/license.md` to `/LICENSE`?

https://github.com/nwutils/nw-builder/blob/90afc6ead4675fa3ae39df98c8c360db2554d7a6/doc/.vitepress/config.js#L20

And also update the link to it in the docs - similar to:

https://github.com/nwutils/nw-builder/blob/90afc6ead4675fa3ae39df98c8c360db2554d7a6/doc/.vitepress/config.js#L13

TheJaredWilcurt:
#893

PR #905, opened 2023-08-02
fix(win): enable icon

hammerOwen:
Fixes: Windows icon not appearing. 

## Description
rcedit is used to modify the windows executable. We need to add the icon option for the icon to be set. Without this modification the executable will never get the icon. 

While testing this locally was seeing weird issues when building the application in the same output directory over and over. 
I moved to using a script to build in different output directories and the builds are entirely consistent now.

It seems that this change originated from this pr #894


ayushmanchhabra:
@hammerOwen Thanks for this! Must have accidently removed it in #894. This is what I see on my end:

Screenshot (5)

I believe the icon not showing in Taskbar is due to https://github.com/nwjs/nw.js/issues/8039

hammerOwen:
Happy to help you have been doing an amazing job supporting this library!

ayushmanchhabra (approved):

ayushmanchhabra:
Holding off on creating a release for now. Found a weird issue introduced in #900

PR #906, opened 2023-08-02
fix(get): return promise to await correctly

hammerOwen:
Fixes: #na

## Description
There is an issue when downloading binaries that the builder doesn't wait for them to download and copies files immediately. 
This causes issues when the target applications files are moved before the file is downloaded. 

To fix this we simply return the promise from the request object. This will then make the function correctly wait. We were seeing intermittent failures on Jenkins when the cache was purged. 

Run without the return 
image

Run with the return
image


ayushmanchhabra (approved):
Thanks! Not sure why Chromedriver test is failing - will fix in a separate PR.

Issue #916, opened 2023-08-10
libffmpeg not being installed into correct folder


zkrige:
### Issue Type

- [X] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour
fetching ffmpeg doesn't seem to install it

### Expected/Proposed Behaviour
Fetching ffmpeg installs it in correct folder

### Additional Info
I'm seeing the original libffmpeg.dylib in the folders on macOS. It looks like its not being overwritten with the one from the community repo

ayushmanchhabra:
You're right, I think I broke this in https://github.com/nwutils/nw-builder/commit/257350d0c00a647342f37fe7494d4121fb4c3bd1

zkrige:
I see this was removed and and is no longer in use

`await replaceFfmpeg(options.platform, nwDir, ffmpegFile);`

It probably just needs to be put back

ayushmanchhabra:
AH yes

zkrige:
Maybe we should keep a ffmpeg folder in the cache dir like you do for nwjs? I think that's how nwjs-builder-phoenix used to do it

zkrige:
Please reopen - this isn't working. I'll sort a fix for this shortly

PR #917, opened 2023-08-10
fix: move ffmpeg to correct folder after extract

zkrige:
Fixes: #916 

## Description
reinstate the `replaceFfmpeg` code that was removed during cleanup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ doc/api*.md
 doc/mode-get.md
 out
 tmp
+.idea
--- a/src/get.js
+++ b/src/get.js
@@ -123,7 +123,7 @@ export async function get({
   }
 
   // If not cached, then download.
-  if (nwCached === false) {
+  if (nwCached === false || ffmpeg === true) {
     log.debug(`Downloading binaries`);
     await mkdir(nwDir, { recursive: true });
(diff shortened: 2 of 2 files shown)


ayushmanchhabra (changes requested):
Thanks for this! One case where this fails is when `ffmpeg` is false and there is a binary cached. Enabling `ffmpeg` does not download it since it thinks its already "cached".

Maybe we can add this [here](https://github.com/nwutils/nw-builder/blob/5726a7ec7ad5fa78e12f53dd9eeb8f58a9c8833d/src/get.js#L125):
```patch
- if (nwCached === false) {
+if (nwCached === false || ffmpeg === true) { 
```

zkrige:
It needs to be a little bit smarter otherwise it will download ffmpeg every time. we only want to download ffmpeg if it hasn't been downloaded.

your solution will for for now, but ideally it should check if ffmpeg must be downloaded

ayushmanchhabra (approved):

zkrige:
@ayushmanchhabra hmmm this is still broken

`await replaceFfmpeg(platform, nwDir, out);`

just ends up copying the zip file over itself. We need some way to get the extracted filename

PR #918, opened 2023-08-11
Copy ffmpeg to correct path after extract

zkrige:
Fixes: #916

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ doc/api*.md
 doc/mode-get.md
 out
 tmp
+.idea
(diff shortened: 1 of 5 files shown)


zkrige:
I haven't verified this on linux or windows, but from what I can see of the folder structure it should be ok

zkrige:
that's odd. how did formatting pass for macOS and linux but not windows?
also prettier doesn't say what's wrong - it just gives a warning

ayushmanchhabra:
That's been failing for other PRs too. Can we be ignored

zkrige:
@ayushmanchhabra could you merge and do a build please. I need this for our CI

ayushmanchhabra:
Could you update the changelog, bump to v4.3.7 and do an `npm install`? That way it would publish on merge.

ayushmanchhabra (approved):

PR #922, opened 2023-08-14
Fix

zkrige:
Fixes: #916 

some versions of node/macOS complain about destination being a file, and others complain when it is only a directory. the only thing I can think to do is to try both

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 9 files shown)


zkrige:
@ayushmanchhabra for some reason the file copy works on my Mac but was failing on circle ci. I've updated the Mac ffmpeg copy to handle both

zkrige:
@ayushmanchhabra would you mind merging this. I need it to get my CI pipeline going

ayushmanchhabra (approved):

Issue #923, opened 2023-08-15
on some macOS, symbolic links are copied as files and link is lost


zkrige:
### Issue Type

- [X] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour
On my circleCI instance symbolic links are lost

### Expected/Proposed Behaviour
links remain

### Additional Info
I'm using node 20.5.0 on my Mac and on CircleCI

on my Mac when use the `get` method my `nwjs-cache/nwjs-v0.78.1-osx-x64/nwjs.app/Contents/Frameworks/nwjs Framework.framework` folder looks like this

```
lrwxr-xr-x@ 1 zayinkrige  staff   24 Aug 15 08:10 Helpers -> Versions/Current/Helpers
lrwxr-xr-x@ 1 zayinkrige  staff   26 Aug 15 08:10 Libraries -> Versions/Current/Libraries
lrwxr-xr-x@ 1 zayinkrige  staff   26 Aug 15 08:10 Resources -> Versions/Current/Resources
drwxr-xr-x@ 4 zayinkrige  staff  128 Aug 15 08:10 Versions
lrwxr-xr-x@ 1 zayinkrige  staff   31 Aug 15 08:10 nwjs Framework -> Versions/Current/nwjs Framework
```

however on CircleCI the same folder looks like this
```
-rwxr-xr-x  1 distiller  staff   24 Aug 15 05:59 Helpers
-rwxr-xr-x  1 distiller  staff   26 Aug 15 05:59 Libraries
-rwxr-xr-x  1 distiller  staff   26 Aug 15 05:59 Resources
drwxr-xr-x  4 distiller  staff  128 Aug 15 05:59 Versions
-rwxr-xr-x  1 distiller  staff   31 Aug 15 05:59 nwjs Framework
```

PR #924, opened 2023-08-15
Fix missing symlinks in unzip

zkrige:
Fixes: #923

## Description
This uses macOS built in `unzip` tool to unzip packages. On some macOS environments, the javascript `compressing` utility isn't restoring symlinks

### DIFF
--- a/src/get.js
+++ b/src/get.js
@@ -164,6 +164,7 @@ export async function get({
                 .uncompress(out, ffmpeg ? nwDir : cacheDir)
                 .then(() => resolve());
             } else if (platform === "osx") {
+              //TODO: compressing package does not restore symlinks on some macOS (eg: circleCI)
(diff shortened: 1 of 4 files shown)


zkrige:
@ayushmanchhabra please let me know if you're ok with this. I'd like to get it merged ASAP

src/get.js
```diff
@@ -162,6 +163,21 @@ export async function get({
               compressing.tgz
                 .uncompress(out, ffmpeg ? nwDir : cacheDir)
                 .then(() => resolve());
+            } else if (platform === "osx") {
```
ayushmanchhabra, line 166: Could you add a comment above this saying something like `TODO: compressing package does not restore symlinks on MacOS`?

ayushmanchhabra (approved):

PR #934, opened 2023-08-24
ci: add Node 16, 18 and 20 [version 3]

ayushmanchhabra:
Fixes: N/A

## Description

By adding more Node versions, we will be able to ensure compatibility across different Node major versions.


ayushmanchhabra:
This PR will have to be broken into smaller parts:
- Fix Jest tests
- Fix Tape tests
- Fix Selenium tests

PR #944, opened 2023-09-04
ffmpeg caching

zkrige:
- Caches ffmpeg
- splits get into get_nwjs and get_ffmpeg

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,6 @@
+node_modules
+
+cache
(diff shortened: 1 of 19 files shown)


ayushmanchhabra:
Looks good except for the test you added (it doesn't really test the caching behaviour). I think we can remove it, merge this in and I'll think of how to test this behaviour (maybe use some fixtures).

PR #956, opened 2023-10-02
fix(types): align type definitions with docs

Enet4:
## Description

This PR changes `index.d.ts` so that:

- The documentation about the various options from index.js are also available through the type definitions;
- Platform-specific app options are also encoded in the types, with type conditionals so that they match the platform selected;
- Optional fields are marked as such, thus preventing errors when they are missing.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,23 +1,175 @@
-type Options = {
-    srcDir: string,
-    mode: "get" | "run" | "build",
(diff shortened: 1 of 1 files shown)


ayushmanchhabra:
@Enet4 Thanks a ton! I've been thinking of adding in depth type definitions but never got around to it. I'll look into this soon!

ayushmanchhabra (approved):

PR #957, opened 2023-10-03
build(jsdoc): autogenerate docs via JSDoc comments

ayushmanchhabra:


src/index.js
```diff
@@ -36,7 +36,46 @@ import { getReleaseInfo } from "./util.js";
  */
 
 /**
- * Automates building an NW.js application.
+ * Installation Guide
+ *
+ * Every NW.js release includes a modified Node.js binary at a specific version. It is recommended to [install](https://nodejs.org/en/download/package-manager) a version greater than or equal to NW.js's Node version. Consult the [version manifest](https://nwjs.io/versions) on the version to install. Install using the package manager of choice. The recommended method is via [corepack](https://nodejs.org/api/corepack.html).
```
TheJaredWilcurt, line 41: You actually want the *exact* version that is built in. For example, if I was using NW.js 0.14.7 which comes with Node 5.11.1. If I'm on Node 19, then when I `npm install` a dependency it may download or build files compatible with Node's ABI based on the global install, which with 19 is 111 but with Node 5, the ABI is 47. So the files downloaded or built will not be compatible with the Node version built in to NW.js.

ayushmanchhabra: Right makes sense. Thanks for catching this - would have definitely overlooked this!

src/index.js
```diff
@@ -36,7 +36,44 @@ import { getReleaseInfo } from "./util.js";
  */
 
 /**
- * Automates building an NW.js application.
+ * Installation Guide
+ *
+ * Every NW.js release includes a modified Node.js binary at a specific version. It is recommended to [install](https://nodejs.org/en/download/package-manager) exactly that version on the host system. Not doing so may download ABI incompatible Node modules. Consult the [version manifest](https://nwjs.io/versions) on the version to install. Install using the package manager of choice via [corepack](https://nodejs.org/api/corepack.html).
```
TheJaredWilcurt, line 41:
```suggestion
 * Every NW.js release includes a modified Node.js binary at a specific version. It is recommended to [install](https://nodejs.org/en/download/package-manager) exactly that version on the host system. Not doing so may download ABI incompatible Node modules. Consult the NW.js [versions manifest](https://nwjs.io/versions) for what Node.js version to install. It is recommended to use a Node version manager (such as [volta](https://volta.sh), n, nvm, or nvm-windows) to be able to easily install and switch between Node versions.
```

PR #932, opened 2023-08-24
build(compressing): unzip via `yauzl-promise` on MacOS 

ayushmanchhabra:
Fixes: #931

## Description

- Removed workaround which requires `unzip` library on MacOS.
- Use `yauzl-promise` to unzip MacOS build on MacOS platform.

## Testing steps

If using in a existing project:

HTTPS:
```shell
npm i -D git+https://github.com/nwutils/nw-builder/tree/dev-931
```

GIT:
```shell
npm i -D git+ssh://git@github.com/nwutils/nw-builder/tree/dev-931
```

Standalone:

1. Clone repo
2. Checkout `dev-931` branch
3. `npm i` to install depencies
4. `npm run demo` to download binaries
5. Check if symlinks are preserved

Issue #931, opened 2023-08-23
Symlinks are copied as files on decompress


ayushmanchhabra:
### Issue Type

- [x] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour

- Symbolic link is copied as a file and the link is lost. More info [here](https://github.com/nwutils/nw-builder/issues/923).

### Expected/Proposed Behaviour

- Symbolic links are recognised.
- Remove `unzip` package from MacOS runner image.

### Additional Info

There is an [PR](https://github.com/node-modules/compressing/pull/31) addressing this issue. Should manually test and see if it resolves issue and speed up review process

Workaround: Do not use symlinks.

- Package version: 4.3.9 onward
- Operating System: MacOS
- Node version: 20.5.0
- NW.js version: 0.79.1

zkrige:
Good work!

I'll test this on our build server asap

zkrige:
Still doesn't work. These are files instead of symlinks

static:nwjs Framework.framework distiller$ ls -al
total 32
drwxr-xr-x 7 distiller staff 224 Aug 25 07:34 .
drwxr-xr-x 3 distiller staff 96 Aug 25 07:34 ..
-rwxr-xr-x 1 distiller staff 24 Aug 25 07:34 Helpers
-rwxr-xr-x 1 distiller staff 26 Aug 25 07:34 Libraries
-rwxr-xr-x 1 distiller staff 26 Aug 25 07:34 Resources
drwxr-xr-x 4 distiller staff 128 Aug 25 07:34 Versions
-rwxr-xr-x 1 distiller staff 31 Aug 25 07:34 nwjs Framework

ayushmanchhabra:
Looking into this

ayushmanchhabra:
@zkrige Could you share your CircleCI env info? I'll add that in to the `compresssing` bug report.

zkrige:
```
Executor / Resource Class
MacOS / M1 Large

declare -x BASH_ENV="/tmp/.bash_env-651a5b86a505136c253fc4f5-0-build"
declare -x BASH_SILENCE_DEPRECATION_WARNING="1"
declare -x CI="true"
declare -x CIRCLECI="true"
declare -x CIRCLE_BRANCH="release"
declare -x CIRCLE_BUILD_NUM="2687"
declare -x CIRCLE_BUILD_URL="<REDACTED>"
[...]

PR #973, opened 2023-11-03
fix(compressing): use `yauzl-promise` if platform and build is MacOS

ayushmanchhabra:
Fixes: #931
Closes: #932

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install packages
         uses: ConorMacBride/install-package@v1.1.0
         with:
-          brew: unzip gpatch
+          brew: gpatch
           choco: patch
           apt: patch
       - name: Checkout repository
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Use `yauzl-promise` to decompress MacOS build on MacOS platform.
 
+### Removed
+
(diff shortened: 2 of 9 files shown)


ayushmanchhabra:
`yauzl-promise` seems to inflate the MacOS app in CI. Selenium tests are the cause of the CI failures.

zkrige:
this is not working on macOS

`await pipeline(readStream, writeStream);` fails with

TypeError [ERR_INVALID_ARG_TYPE]: The "streams[stream.length - 1]" property must be of type function. Received an instance of WriteStream
    at popCallback (node:internal/streams/pipeline:74:3)
    at pipeline (node:internal/streams/pipeline:185:37)
    at file:....node_modules/nw-builder/src/get.js:253:19
    at get (file:....node_modules/nw-builder/src/get.js:90:3)
    at Module.nwbuild (file:....node_modules/nw-builder/src/index.js:149:5)

zkrige:
my fix gets around the error, but it doesnt solve the root cause - symlinks are still extracted as files

```
Zayins-MacBook-Pro:Versions zayinkrige 11:45:27$ ls -al
total 6272
drwxr-xr-x@ 4 zayinkrige  staff      128 Nov  9 11:44 .
drwxr-xr-x@ 7 zayinkrige  staff      224 Nov  9 11:44 ..
drwxr-xr-x@ 8 zayinkrige  staff      256 Nov  9 11:44 119.0.6045.105
-rw-r--r--@ 1 zayinkrige  staff  2868248 Nov  9 11:44 Current

```

PR #975, opened 2023-11-09
fix for macOS unzip

zkrige:
Fixes: #973


src/get.js
```diff
@@ -243,16 +241,29 @@ async function getNwjs({
...
+            await mkdir(directory, { recursive: true });
+
             const readStream = await entry.openReadStream();
-            const writeStream = createWriteStream(
-              `${cacheDir}/${entry.filename}`,
-            );
-            await pipeline(readStream, writeStream);
+            const writeStream = createWriteStream(fullEntryPath);
+
+            await new Promise((resolve, reject) => {
```
ayushmanchhabra, line 257: Is this necessary? Wouldn't the error be caught by the catch block?

zkrige:
Thanks for merging this - but it still doesnt fix the symlink issue :)

PR #989, opened 2023-12-04
fix(ci): stream handling for macos

ayushmanchhabra:
Fixes: N/A

## Description

- Use Stream Promise API with yauzl-promise.

### DIFF
--- a/src/get.js
+++ b/src/get.js
@@ -13,6 +13,7 @@ import yauzl from "yauzl-promise";
 import util from "./util.js";
 
 import "./nwbuild.js";
+import streamp from "node:stream/promises";
 
 /**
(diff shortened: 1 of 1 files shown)


ayushmanchhabra:
https://github.com/nwutils/nw-builder/actions/runs/7091596162/job/19301077432
@sysrage @zkrige Not sure why the stream is closing prematurely. Tested locally on Ubuntu 22 and Windows 11 and don't see this error happening. Could you verify if it happens on your local MacOS?

zkrige:
I'm not sure why its closing early, but we should move away from yauzl for macOS anyways - it doesnt extract symlinks, so theres no point in keeping it

PR #999, opened 2023-12-21
fix(get): wrap `unzipper` in Promise

jssuttles:
## Description
Before this change, it didn't wait for the unzipping to finish before letting the next part of the process happen, the build. This breaks using this package within a build process because the unzipping doesn't finish before the build expects to use the unzipped folder. This is a race condition.


ayushmanchhabra (approved):
Thank you!

PR #1008, opened 2024-01-16
feat(chore): cleanup and optimize import

jorisdugue:
Fixes: N/A

## Description

Cleanup this global import for reduce memory in build 😄


ayushmanchhabra:
Is there a significant change in memory usage by changing from default to regular imports? Builds currently take time since it removes the existing cached binaries and decompresses the downloaded file.

CODE_OF_CONDUCT.md
```diff
@@ -46,9 +46,9 @@ Any discussion or communication that is primarily focused around an ideology, be
...
+2. If a member repeatedly violates the guidelines established in this document, they should be politely warned that continuing to violate the rules may result in being banned from the community. This means revoking access and support to interactions relating directly to the project (issue trackers, chatrooms, forums, in person events, etc.). However, they may continue to use the technology in accordance with its license.
+3. If a maintainer is in violation of a guideline, they should be informed of such with a link to this document. If additional actions are required of the maintainer but not taken, then other maintainers should be informed of these inactions.
+4. If a maintainer repeatedly violates the guidelines established in this document, they should be politely warned that continuing to violate the rules may result in being banned from the community. This means revoking access and support to interactions relating directly to the project (issue trackers, chatrooms, forums, in person events, etc.). However, they may continue to use the technology in accordance with its license. In addition, future contributions to this project may be ignored as well.
```
TheJaredWilcurt, line 51:
```suggestion
1. If a member repeatedly violates the guidelines established in this document, they should be politely warned that continuing to violate the rules may result in being banned from the community. This means revoking access and support to interactions relating directly to the project (issue trackers, chatrooms, forums, in person events, etc.). However, they may continue to use the technology in accordance with its license.
1. If a maintainer is in violation of a guideline, they should be informed of such with a link to this document. If additional actions are required of the maintainer but not taken, then other maintainers should be informed of these inactions.
1. If a maintainer repeatedly violates the guidelines established in this document, they should be politely warned that continuing to violate the rules may result in being banned from the community. This means revoking access and support to interactions relating directly to the project (issue trackers, chatrooms, forums, in person events, etc.). However, they may continue to use the technology in accordance with its license. In addition, future contributions to this project may be ignored as well.
```

src/bld.js
```diff
@@ -270,7 +270,7 @@ const manageManifest = async ({ nwPkg, managedManifest, outDir, platform }) => {
   );
 
   process.chdir(
-    path.resolve(
```
TheJaredWilcurt, line 273: This looks like premature optimization. Unless you can measure the actual benefits of this and they are substantial, I don't think it justifies reducing code readability. For example `resolve` is a word commonly associated to promises. `(resolve, reject)`. So seeing it by itself I would assume that is what it relates to. However, seeing it attached to the `path` import, I know that it actually gives me a resolved file path. This goes with all the rest, seeing the associated hierarchy of the import is generally helpful while working on a code base.

Readability and maintainability are more important than build time or memory usage.

What impact would the performance increase have? If it only shaves off a few seconds of a multi-minute process, it's not worth doing.

I wouldn't mind faster builds, but I think we would need to focus on actual architectural and 3rd-party dependency evaluation over saving a few KB of memory by skipping importing a native node module.

jorisdugue: This certainly saves a few kb on launch but during use it has an effect allowing you to save a few seconds. I can understand that this is a small impact after all depends on the number of builds you launch in a row 😄 

For the Resolve part I was careful not to end up in promises that could have posed a potential conflict

ayushmanchhabra (commented):
@jorisdugue Thanks for the PR but I won't be accepting it. I generally agree with what Jared said "Readability and maintainability are more important than build time or memory usage."
I'm in the process of improving documentation, contributing guide and testing. Hopefully it becomes easier for people to contribute to this project in the future.

PR #1025, opened 2024-02-02
chore(get): refactor implementation

ayushmanchhabra:
### Description of Changes

- Use internal APIs such as request, decompress, ffmpeg, node and nw functions.


ayushmanchhabra:
This actually fixes cache behaviour but the commit message does not capture this info.

Refs: #1037

PR #1036, opened 2024-02-15
fix(bld): build with `options.managedManifest` when `options.glob` is disabled

ayushmanchhabra:
*


ayushmanchhabra:
Revisit later. Will need to decide and document if `options.managedManifest` not working with `options.glob` disabled is a bug or a design limitation.

PR #1040, opened 2024-02-21
Fixes

zkrige:
I tried the latest version on macOS and a lot of things broke.

Here are the fixes

- fix lint scripts
- fix macos symlinks
- fix ffmpeg download

### DIFF
--- a/src/get/decompress.js
+++ b/src/get/decompress.js
@@ -26,7 +26,7 @@ export default async function decompress(filePath, cacheDir) {
  * Get file mode from entry. Reference implementation is [here](https://github.com/fpsqdb/zip-lib/blob/ac447d269218d396e05cd7072d0e9cd82b5ec52c/src/unzip.ts#L380).
  *
  * @param  {yauzl.Entry} entry  - Yauzl entry
- * @return {number}             - entry's file mode
+ * @return {number}             - entry's file mode 
  */
 function modeFromEntry(entry) {
   const attr = entry.externalFileAttributes >> 16 || 33188;
--- a/src/get/decompress.js
+++ b/src/get/decompress.js
@@ -26,7 +26,7 @@ export default async function decompress(filePath, cacheDir) {
  * Get file mode from entry. Reference implementation is [here](https://github.com/fpsqdb/zip-lib/blob/ac447d269218d396e05cd7072d0e9cd82b5ec52c/src/unzip.ts#L380).
  *
  * @param  {yauzl.Entry} entry  - Yauzl entry
- * @return {number}             - entry's file mode 
+ * @return {number}             - entry's file mode
  */
 function modeFromEntry(entry) {
   const attr = entry.externalFileAttributes >> 16 || 33188;
@@ -47,31 +47,50 @@ function modeFromEntry(entry) {
  */
 async function unzip(zippedFile, cacheDir) {
   const zip = await yauzl.open(zippedFile);
-
   let entry = await zip.readEntry();
+  const symlinks = []; // Array to hold symbolic link entries
 
   while (entry !== null) {
     let entryPathAbs = path.join(cacheDir, entry.filename);
-    /* Create the directory beforehand to prevent `ENOENT: no such file or directory` errors. */
-    await fs.promises.mkdir(path.dirname(entryPathAbs), { recursive: true });
-    /* Check if entry is a symbolic link */
+    // Check if entry is a symbolic link
     const isSymlink = ((modeFromEntry(entry) & 0o170000) === 0o120000);
-    const readStream = await entry.openReadStream();
-    
+
     if (isSymlink) {
-      const chunks = [];
-      /* Read stream into Array. */
-      readStream.on("data", (chunk) => chunks.push(chunk));
-      await stream.promises.finished(readStream);
-      const link = Buffer.concat(chunks).toString('utf8').trim();
-      await fs.promises.symlink(link, entryPathAbs)
+      // Store symlink entries to process later
+      symlinks.push(entry);
     } else {
-      // Pipe read to write stream
-      const writeStream = fs.createWriteStream(entryPathAbs);
-      await stream.promises.pipeline(readStream, writeStream);
+      // Handle regular files and directories
+      await fs.promises.mkdir(path.dirname(entryPathAbs), {recursive: true});
+      if (!entry.filename.endsWith('/')) { // Skip directories
+        const readStream = await entry.openReadStream();
+        const writeStream = fs.createWriteStream(entryPathAbs);
+        await stream.promises.pipeline(readStream, writeStream);
+
+        // Set file permissions after the file has been written
+        const mode = modeFromEntry(entry);
+        await fs.promises.chmod(entryPathAbs, mode);
+      }
     }
 
     // Read next entry
     entry = await zip.readEntry();
+  }
+
+  // Process symbolic links after all other files have been extracted
+  for (const symlinkEntry of symlinks) {
+    let entryPathAbs = path.join(cacheDir, symlinkEntry.filename);
+    const readStream = await symlinkEntry.openReadStream();
+    const chunks = [];
+    readStream.on("data", (chunk) => chunks.push(chunk));
+    await new Promise(resolve => readStream.on("end", resolve));
+    const linkTarget = Buffer.concat(chunks).toString('utf8').trim();
+
+    // Check if the symlink or a file/directory already exists at the destination
+    if (fs.existsSync(entryPathAbs)) {
+      //skip
+    } else {
+      // Create symbolic link
+      await fs.promises.symlink(linkTarget, entryPathAbs);
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


src/get/index.js
```diff
@@ -191,7 +193,7 @@ async function get(options) {
     }
 
     /**
-     * If the compressed binary exists, then `true`. Otherwise, it is `false`. 
+     * If the compressed binary exists, then `true`. Otherwise, it is `false`.
```
ayushmanchhabra, line 196:
```suggestion
     * If the compressed binary exists, then `true`. Otherwise, it is `false`. 
```

src/get/index.js
```diff
@@ -109,13 +109,15 @@ async function get(options) {
...
      *
      * @type {boolean}
      */
     const ffmpegFilePathExists = await util.fileExists(ffmpegFilePath);
     if (ffmpegFilePathExists === false) {
-      ffmpegFilePath = await ffmpeg(options.downloadUrl, options.version, options.platform, options.arch, options.cacheDir);
+      const downloadUrl =
+        "https://github.com/nwjs-ffmpeg-prebuilt/nwjs-ffmpeg-prebuilt/releases/download";
+      ffmpegFilePath = await ffmpeg(downloadUrl, options.version, options.platform, options.arch, options.cacheDir);
```
ayushmanchhabra, line 120:
```suggestion
      options.downloadUrl =
        "https://github.com/nwjs-ffmpeg-prebuilt/nwjs-ffmpeg-prebuilt/releases/download";
      ffmpegFilePath = await ffmpeg(options.downloadUrl, options.version, options.platform, options.arch, options.cacheDir);
```

src/get/index.js
```diff
@@ -97,7 +97,7 @@ async function get(options) {
      */
     let ffmpegFilePath = path.resolve(
       options.cacheDir,
-      `nwjs${options.flavor === "sdk" ? "-sdk" : ""}-v${options.version}-${options.platform}-${options.arch}.zip`,
+      `ffmpeg-${options.version}-${options.platform}-${options.arch}.zip`,
```
ayushmanchhabra, line 100: Thanks for this! Must have missed this while refactoring.

src/get/index.js
```diff
@@ -74,10 +74,10 @@ async function get(options) {
   // but want a subsequent build with ffmpeg flag disabled. By removing the directory and
   // decompressing it again, we prevent the community ffmpeg files from being left over.
   // This is important since the community ffmpeg builds have specific licensing constraints.
-  await fs.promises.rm(nwDirPath, { recursive: true, force: true });
+  await fs.promises.rm(nwDirPath, {recursive: true, force: true});
 
   /**
-   * If the compressed binary exists, then `true`. Otherwise, it is `false`. 
+   * If the compressed binary exists, then `true`. Otherwise, it is `false`.
```
ayushmanchhabra, line 80:
```suggestion
   * If the compressed binary exists, then `true`. Otherwise, it is `false`. 
```

src/get/index.js
```diff
@@ -74,10 +74,10 @@ async function get(options) {
   // but want a subsequent build with ffmpeg flag disabled. By removing the directory and
   // decompressing it again, we prevent the community ffmpeg files from being left over.
   // This is important since the community ffmpeg builds have specific licensing constraints.
-  await fs.promises.rm(nwDirPath, { recursive: true, force: true });
+  await fs.promises.rm(nwDirPath, {recursive: true, force: true});
```
ayushmanchhabra, line 77:
```suggestion
  await fs.promises.rm(nwDirPath, { recursive: true, force: true });
```

src/get/index.js
```diff
@@ -32,13 +32,13 @@ import util from "../util.js";
...
   /**
-   * If `options.cacheDir` exists, then `true`. Otherwise, it is `false`. 
+   * If `options.cacheDir` exists, then `true`. Otherwise, it is `false`.
    *
    * @type {boolean}
    */
   const cacheDirExists = await util.fileExists(options.cacheDir);
   if (cacheDirExists === false) {
-    await fs.promises.mkdir(options.cacheDir, { recursive: true });
+    await fs.promises.mkdir(options.cacheDir, {recursive: true});
```
ayushmanchhabra, line 41:
```suggestion
    await fs.promises.mkdir(options.cacheDir, { recursive: true });
```

src/get/index.js
```diff
@@ -32,13 +32,13 @@ import util from "../util.js";
 async function get(options) {
 
   /**
-   * If `options.cacheDir` exists, then `true`. Otherwise, it is `false`. 
+   * If `options.cacheDir` exists, then `true`. Otherwise, it is `false`.
```
ayushmanchhabra, line 35:
```suggestion
   * If `options.cacheDir` exists, then `true`. Otherwise, it is `false`. 
```

src/get/decompress.js
```diff
@@ -26,7 +26,7 @@ export default async function decompress(filePath, cacheDir) {
  * Get file mode from entry. Reference implementation is [here](https://github.com/fpsqdb/zip-lib/blob/ac447d269218d396e05cd7072d0e9cd82b5ec52c/src/unzip.ts#L380).
  *
  * @param  {yauzl.Entry} entry  - Yauzl entry
- * @return {number}             - entry's file mode 
+ * @return {number}             - entry's file mode
```
ayushmanchhabra, line 29:
```suggestion
 * @return {number}             - entry's file mode 
```

src/get/index.js
zkrige: I'm not sure it's necessary to set the downloadurl for ffmpeg onto the global options. Remember this is JUST the ffmpeg download url, not the nwjs download url

ayushmanchhabra: Aghh true, nice catch! This would probably error out when `options.ffmpeg` _and_ `options.nativeAddon` are enabled.

src/get/index.js
```diff
@@ -115,7 +115,9 @@ async function get(options) {
      */
     const ffmpegFilePathExists = await util.fileExists(ffmpegFilePath);
     if (ffmpegFilePathExists === false) {
-      ffmpegFilePath = await ffmpeg(options.downloadUrl, options.version, options.platform, options.arch, options.cacheDir);
+      const downloadUrl =
```
ayushmanchhabra, line 118:
```suggestion
     // Do not update the options.downloadUrl with the ffmpeg URL here. Doing so would lead to error when options.ffmpeg and options.nativeAddon are both enabled.
      const downloadUrl =
```

ayushmanchhabra: This should prevent my future self from changing this behaviour

src/get/decompress.js
```diff
@@ -47,31 +47,50 @@ function modeFromEntry(entry) {
...
+        await fs.promises.chmod(entryPathAbs, mode);
+      }
     }
 
     // Read next entry
     entry = await zip.readEntry();
   }
+
+  // Process symbolic links after all other files have been extracted
+  for (const symlinkEntry of symlinks) {
```
ayushmanchhabra, line 80: This is essentially how symlinks were handled before #1035. How does the current implementation fail? And can it be covered by a test?

ayushmanchhabra:
> Is there anything else needed? Must I change version number or anything?

Nope, releases have been automated now. One last question and should be good to go.

src/get/decompress.js
```diff
@@ -47,31 +47,50 @@ function modeFromEntry(entry) {
...
+    let entryPathAbs = path.join(cacheDir, symlinkEntry.filename);
+    const readStream = await symlinkEntry.openReadStream();
+    const chunks = [];
+    readStream.on("data", (chunk) => chunks.push(chunk));
+    await new Promise(resolve => readStream.on("end", resolve));
+    const linkTarget = Buffer.concat(chunks).toString('utf8').trim();
+
+    // Check if the symlink or a file/directory already exists at the destination
+    if (fs.existsSync(entryPathAbs)) {
```
ayushmanchhabra, line 89: Does [`rm`](https://github.com/nwutils/nw-builder/blob/2284d52ba0c4d1bb9a86a91e138ba99ce2af8d6c/src/get/index.js#L77) not delete symlinks? Why should there be existing symlinks?

src/get/decompress.js
zkrige: The current implementation just doesnt extract the symlinks. The new implementation first extracts all the files and then creates symlinks after. This new method is the same method that macOS unzip uses - see attached screenshot
image

ayushmanchhabra (approved):

zkrige:
@ayushmanchhabra I merged this PR but I see version is still 4.6.1 - should I manually bump version?

ayushmanchhabra:
Yup thats cuz you merged instead of squash and merge, let me try squashing the last few commits

zkrige:
ah apologies - I'll bear that in mind for futuer

ayushmanchhabra:
All good, a simple squash and merge and voila!

PR #1044, opened 2024-02-24
fix(get): copy ffmpeg to correct location on windows

zkrige:
* Windows FFmpeg copy was breaking because the destination path was never set


zkrige:
this is tested working on windows builds. Previous it was failing on windows with error

ayushmanchhabra (approved):

PR #1094, opened 2024-05-16
chore(bld): Replace rcedit with resedit

CosmoMyzrailGorynych:
* Replaces `rcedit` dependency with `resedit`

Why?

* `rcedit` is quite an ancient package that requires Wine to do anything on Linux/Windows, and having Wine as a dependency in Linux workers sucks!
* `resedit` is instead a vanilla-node.js package that does not require building or external binaries.

Replacing `rcedit` with `resedit` allows cross-building Nw.js apps for Windows on mainstream Linux+Node.js container images. (Previously you would make your own to preinstall Wine.) And, of course, it makes nw-builder easier to use on Linux-based personal machines, too!


CosmoMyzrailGorynych:
Why would an option be needed if no API changes were introduced?

вс, 19 мая 2024 г., 19:57 Ayushman Chhabra ***@***.***>:

CosmoMyzrailGorynych:
> Thanks for this. I can see why you'd want to replace rcedit with resedit but doing so would introduce a breaking change to current developers workflow.

Please explain the breaking changes here. There are no API changes and no new dependencies.

TheJaredWilcurt:
* https://www.npmjs.com/package/resedit

The resedit package makes no claims of being a drop-in replacement for, or even having API parity with rcedit. It in fact does not even mention rcedit *at all*. So it would be safe to assume some form of breaking changes would be introduced.

> rcedit is quite ancient

Both rcedit and resedit are actively maintained and have had new releases in the past year (November and April).

I agree with Ayush. Offering both as an option makes the most sense. Continue to default to what we have to avoid breaking changes. Re-evaluate when going to the next major in the library.

I'd like to see a more comprehensive comparison of both technologies, what features they each have, if there are any features one has the other doesn't, are there granular controls that differ, etc (other than wine which is the obvious major difference). If we could completely replace rcedit, that would be cool, I am generally in favor of tools that are cross-platform by default and don't require the dev to jump through hoops like setting up wine, or using a separate Windows VM for builds.

CosmoMyzrailGorynych:
> The resedit package makes no claims of being a drop-in replacement for, or even having API parity with rcedit.

They don't — my PR handles these differences so the end product is the same. `nw-builder` uses `rcedit` to change executable's main icon and change the version info (which includes lots of other keys like company name and copyright), and so does `resedit` in my PR. Icons are replaced in a correct icon group. List of keys Windows executables can have is neither up to `rcedit` nor up to `resedit` — it is constant and thus is the same for both packages.

I use `resedit` for my nw.js app and for neutralino.js ones, `resedit` does a great job in both circumstances. Disputes about keeping `rcedit` seems like an XY problem to me or not understanding the PR.

src/bld.js
```diff
@@ -311,28 +313,33 @@ const setWinConfig = async ({ app, outDir }) => {
...
-  } catch (error) {
-    if (process.platform !== "win32") {
-      console.warn(
-        "Ensure WINE is installed or build your application on Windows platform",
-      );
-    }
-    throw error;
+    const iconBuffer = await fsm.readFile(path.resolve(app.icon));
+    const iconFile = resedit.Data.IconFile.from(iconBuffer);
+    resedit.Resource.IconGroupEntry.replaceIconsForResource(
```
TheJaredWilcurt, line 323:
```suggestion
    // English (United States)
    const EN_US = 1033;
    resedit.Resource.IconGroupEntry.replaceIconsForResource(
```

src/bld.js
```diff
@@ -311,28 +313,33 @@ const setWinConfig = async ({ app, outDir }) => {
...
+      res.entries,
+      // This is the name of the icon group nw.js uses that gets shown in file exlorers
+      'IDR_MAINFRAME',
+      // 1033 means "English (United States)"
+      1033,
```
TheJaredWilcurt, line 328:
```suggestion
      EN_US,
```

src/bld.js
```diff
@@ -311,28 +313,33 @@ const setWinConfig = async ({ app, outDir }) => {
...
+      iconFile.icons.map(i => i.data)
+    );
   }
+  const [vi] = resedit.Resource.VersionInfo.fromEntries(res.entries);
+  const [major, minor, patch] = app.version.split(".");
+  vi.setFileVersion(major, minor, patch, 0, 1033);
```
TheJaredWilcurt, line 334:
```suggestion
  vi.setFileVersion(major, minor, patch, 0, EN_US);
```

src/bld.js
```diff
@@ -311,28 +313,33 @@ const setWinConfig = async ({ app, outDir }) => {
...
+  vi.setStringValues({
+    lang: 1033,
```
TheJaredWilcurt, line 336:
```suggestion
    lang: EN_US,
```

CosmoMyzrailGorynych:
Thank you for merging in the review changes and the PR — I was busy with my own projects and couldn't add changes in time.

Issue #1131, opened 2024-06-22
postinstall script fails because of missing base-volta-off-of-nwjs


thisjt:
### Issue Type

- [x] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour

I am getting this error when I attempt to update to 4.7.6.
![image](https://github.com/nwutils/nw-builder/assets/64672207/7596b268-9d8a-4185-9379-0dbb2e5576fd)

### Expected/Proposed Behaviour

postinstall script will run without issues after updating

### Additional Info

- Package version: 4.7.5/6
- Operating System: Windows 11
- Node version: 20.13.1
- NW.js version: 0.88.0-sdk
- Repro link: create a package.json and do `pnpm i -D nw-builder@latest`

TheJaredWilcurt:
This is because you are installing it as a devDependency, and devDependencies do not install *their* devDependency, so it is trying to run a post-install script referencing a node module that was skipped.

I designed `base-volta-off-of-nwjs` to be ran via `npx`, so this kind of thing wouldn't happen, but also, I hadn't considered its use with a library, rather than a desktop app.

Perhaps changing to a different npm hook would work better. I don't know which one (if any) is better though:

* https://docs.npmjs.com/cli/v10/using-npm/scripts

@ayushmanchhabra You'd want something that runs locally for development, but isn't ran when installed as a dependency when published.

TheJaredWilcurt:
This repo could create a `postinstall.js` script that checks if `node_modules/.bin/base-volta-off-of-nwjs` exists, and if so, runs it. Then change the `package.json` to `"postinstall": "node ./scripts/postinstall.js"`. That's the only solution I can think of to work around this.

thisjt:
Confirmed fixed in 4.7.7. Thanks!
![image](https://github.com/nwutils/nw-builder/assets/64672207/88f80ae0-93a4-4052-930a-b88f93100332)

PR #1132, opened 2024-06-22
Create postinstall script for basing volta of nw.js

TheJaredWilcurt:
Fixes: #1131

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,0 +1,12 @@
+import { execSync } from 'node:child_process';
+import { existsSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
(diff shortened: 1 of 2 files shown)


scripts/postinstall.js
```diff
@@ -0,0 +1,12 @@
+import { execSync } from 'node:child_process';
+import { existsSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+const baseVoltaPath = resolve(join(__dirname, '..', 'node_modules', 'base-volta-off-of-nwjs', 'index.js'));
+
+if (existsSync(baseVoltaPath)) {
```
TheJaredWilcurt, line 10:
```suggestion
/**
 * When developing locally, we want to make sure the Node.js version
 * documented in the `volta` object in the `package.json` matches
 * whatever version of NW.js is installed, for manual testing purposes.
 * However, for those installing this library as a devDependency, they
 * will not have `base-volta-off-of-nwjs` installed, and nor should
 * they. So we skip this check for them.
 */
if (existsSync(baseVoltaPath)) {
```

Issue #1235, opened 2024-09-20
New mac-specific option needed: `NSLocalNetworkUsageDescription`


beaugunderson:
### Issue Type

- [ ] Bug Report
- [x] Feature Request
- [ ] Other

### Current/Missing Behaviour

macOS 15 now requires `NSLocalNetworkUsageDescription` in the `Info.plist` to enable local networking, so we need to be able to specify it.

### Expected/Proposed Behaviour

When passed to `app` for a macOS build, the `NSLocalNetworkUsageDescription` key is added to `Info.plist` like the other keys.

### Additional Info

- Package version: latest
- Operating System: macOS 15
- Node version: latest
- NW.js version: latest
- ...

ayushmanchhabra:
https://developer.apple.com/documentation/bundleresources/information_property_list/nslocalnetworkusagedescription
@beaugunderson Looks like it is supported since MacOS 11. Are you open to creating a pull request?

beaugunderson:
There's a FAQ here as well: https://developer.apple.com/forums/thread/663858

beaugunderson:
Very open! I'll do that now. :)

beaugunderson:
Done!

PR #1236, opened 2024-09-20
feat(bld): add NSLocalNetworkUsageDescription option for MacOS

beaugunderson:
* Declare why an app needs to access the local network

Fixes: #1235

Refs: https://developer.apple.com/forums/thread/663858
Refs: https://developer.apple.com/documentation/bundleresources/information_property_list/nslocalnetworkusagedescription


ayushmanchhabra (approved):
Thanks!

Issue #1258, opened 2024-10-07
app.name may not be filesafe


Black-Platypus:
### Issue Type

- [x] Bug Report
- [ ] Feature Request
- [ ] Other

### Current/Missing Behaviour
When using an app name that includes illegal characters for filenames, renaming nw.exe (etc) may fail or lead to unintended behavior (also maybe path traversal?)
### Expected/Proposed Behaviour
The executable is renamed taking the above into account
### Additional Info

- Package version: 4.11.3
- Operating System: Windows 10 x64
- Node version: v20.9.0, 22.7.0
- NW.js version: 0.92.0

___
Depending on whether the final executable path needs to be known after it was created or not, I think there should be a remedy either in the parsing stage or the renaming stage.
I would encourage getting everything prepared ahead of taking any action, so I'd like to suggest ideally having a theoretical stage where everything is prepared, or at least making sure the `app.name` is file safe in the parsing stage, to minimize issues arising from potential differences between `app.name` and the resulting executable name.

Black-Platypus:
PRs:
Parsing stage: #1259 
Renaming stage (windows): #1260

ayushmanchhabra:
@Black-Platypus Thanks for the PRs, I'll try to review them over next couple days

PR #1259, opened 2024-10-07
fix(util): strip special and control characters from app.name

Black-Platypus:
* This mitigates potential path traversal.

Fixes: #1258

PR #1260, opened 2024-10-07
Windows: Make executable name filesafe

Black-Platypus:
Windows: Made renaming filesafe in the renaming stage

Fixes: #1258


ayushmanchhabra:
Is this still needed? In #1259, the special and control characters have been stripped from app.name.

Black-Platypus:
No, if you're going with #1259, this is not needed (I just provided it as an alternative, since I can only really speak to Windows)

Issue #1261, opened 2024-10-07
Don't discard information previously set by user when some nwbuilder manifest is found


Black-Platypus:
### Issue Type

- [ ] Bug Report
- [x] Feature Request
- [ ] Other

### Current/Missing Behaviour
User passes in an options Object for a project that has a "nwbuild" object in its package/manifest JSON.
All of the previously entered information is discarded, regardless of what's inside the `nwbuild` object
### Expected/Proposed Behaviour
Keys previously explicitly specified in the options object when calling `nwbuilder.bld` remain when they're not overwritten by any manifest encountered
### Additional Info

- Package version: 4.11.3
- Operating System: Windows 10 x64
- Node version: v20.9.0, 22.7.0
- NW.js version: 0.92.0

ayushmanchhabra:
Didn't read through this well enough earlier. Maybe this should be documented somewhere.

PR #1262, opened 2024-10-07
Don't overwrite all options previously specified when manifest is found (assign re-defined only)

Black-Platypus:
Changed behavior when finding a possible existing nwbuild manifest:  
Don't overwrite all options previously specified when manifest is found (assign re-defined only)

Fixes: #1261


Black-Platypus:
(Sorry, wrong history; See instead #1263)

PR #1263, opened 2024-10-07
Don't overwrite all options previously specified when manifest is found (assign re-defined only)

Black-Platypus:
Keep anything the user specified in the `options` object and only overwrite things found in any manifest (package.json) found.
- Assign `manifest.app` to `options.app`
- Assign `manifest` to `options`
- Restore `options.app`

Fixes: #1261

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -43,11 +43,15 @@ async function nwbuild(options) {
 
   try {
     // Parse options
-    options = await util.parse(options, manifest);
+    // options = await util.parse(options, manifest); // Preserve user input for now, assign manifest overwrites, only then parse
 
     manifest = await util.getNodeManifest({ srcDir: options.srcDir, glob: options.glob });
     if (typeof manifest?.nwbuild === 'object') {
-      options = manifest.nwbuild;
+      if(typeof manifest.nwbuild.app === 'object')
+        Object.assign(options.app, manifest.nwbuild.app);
+      let appOptions = options.app;
+      Object.assign(options, manifest.nwbuild);
+      options.app = appOptions;
     }
 
     options = await util.parse(options, manifest);


src/index.js
```diff
@@ -43,11 +43,15 @@ async function nwbuild(options) {
 
   try {
     // Parse options
-    options = await util.parse(options, manifest);
```
ayushmanchhabra, line 46: I remember why I did this. If options is empty, then getNodeManifest function will throw an error.

ayushmanchhabra: @Black-Platypus ping

Black-Platypus:
> > Keep anything the user specified in the `options` object and only overwrite things found in any manifest (package.json) found.
> 
> Also nothing should be overwritten. If an option is undefined, the default value is used for it. If an option is being overwritten, then its probably an error in the parse function.

This completely overwrites `options`, no? Am I overlooking something?
https://github.com/nwutils/nw-builder/blob/ef120b20b097f2706939864f7e9a62a665902342/src/index.js#L50

ayushmanchhabra:
> > > Keep anything the user specified in the `options` object and only overwrite things found in any manifest (package.json) found.
> > 
> > 
> > Also nothing should be overwritten. If an option is undefined, the default value is used for it. If an option is being overwritten, then its probably an error in the parse function.
> 
> This completely overwrites `options`, no? Am I overlooking something?
> 
> https://github.com/nwutils/nw-builder/blob/ef120b20b097f2706939864f7e9a62a665902342/src/index.js#L50

The options defined in a package.json (NW.js manifest) overwrite options defined in JavaScript API or CLI. This is a feature not bug.

Black-Platypus:
Alright, sad to hear this won't be implemented.
(It felt wrong to discard information specifically set in JS or CLI just because I want to pass in _some_ values via manifest)

Black-Platypus:
Alright, sad to hear this won't be implemented.
(It felt wrong to discard information specifically set in JS or CLI just because I want to pass in some values via manifest)

ayushmanchhabra:
I'm not against this but it would be a breaking change. Also having a priority makes it easier to maintain in the long term

Issue #1264, opened 2024-10-07
Provide calculated values (final paths, names, versions, etc) to user without building app


Black-Platypus:
### Issue Type

- [ ] Bug Report
- [x] Feature Request
- [ ] Other

### Current/Missing Behaviour
User has their values they pass to nwbuild, but no way of double-checking, changing or even seeing any value that nw-builder arrives at
### Expected/Proposed Behaviour
User has an alternate "prepare" mode, where they can pass in their options and receive the calculated state (after parsing and applying defaults and path resolutions), to be able to check for validity and/or make adjustments
Usage like:
```
nwbuild({
	mode: 'prepare'
	// ...
});
```

### Additional Info

- Package version: 4.11.3
- Operating System: Windows 10 x64
- Node version: v20.9.0, 22.7.0
- NW.js version: 0.92.0

ayushmanchhabra:
https://github.com/nwutils/nw-builder/pull/1265#issuecomment-2488425920

PR #1265, opened 2024-10-07
Implement "prepare" mode: Let the user see and work with the calculated values (final paths, names, versions, etc) without having to build and inspect the results

Black-Platypus:
Add "prepare" mode, calculating all major values and returning it to user, without having to "build" anything.
Use output object to build if `options.mode` is `'build'`
This will allow for builder apps to better handle building apps using `nw-builder` without having to know in advance what the nwjs app is going to provide.
Using things like, for example,
```
var meta = await nwbuild({
	mode: 'prepare'
	// ...
});
if( !meta.app.legalCopyright && meta.app.company){
	meta.app.legalCopyright = "© " + meta.app.company + " " + ((new Date()).getFullYear());
}
// ...
// Now that we've done our changes, build using the previously arrived-at values
meta.mode = 'build';
await nwbuild(meta);
```
, a lot of things can be streamlined, without having to parse the app's manifest itself, without knowing in advance how nw-builder will parse/assign values, without having to make sure that any path resolves to what the builder app thinks/hopes it does (hopefully it should just be resolving things within the CWD, but who knows), etc

**ToDo**: Implement OSX behavior (has its own file, I'm not sure of the reasons, nor can I test the behavior)

Fixes: #1264

### DIFF
--- a/src/bld.js
+++ b/src/bld.js
@@ -122,44 +122,65 @@ async function bld({
   nativeAddon = false,
   zip = false,
   releaseInfo = {},
+  mode = 'build'
 }) {
   const nwDir = path.resolve(
     cacheDir,
     `nwjs${flavor === 'sdk' ? '-sdk' : ''}-v${version}-${platform
     }-${arch}`,
   );
+  
+  if(mode == "build"){
+    // TODO: Check if outDir is really a previous built dir (known files?) to avoid catastrophy on misconfiguration
+    await fs.promises.rm(outDir, { force: true, recursive: true });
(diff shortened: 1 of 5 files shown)


ayushmanchhabra:
I had thought of something similar to this a while back but dropped the idea because I wanted to keep things [as open as possible](https://en.wikipedia.org/wiki/Inversion_of_control). Will review this last since this is a feature request.

ayushmanchhabra:
@Black-Platypus I have implemented logging through which you are able to see what the values of the options are. You will hve to set `options.quiet` to `debug`. Does that fit your use case? (Apologies for the late reply, somehow missed this)

PR #1304, opened 2024-11-20
feat(run): return NW.js process reference

ayushmanchhabra:
* This allows us to kill the process via code which is useful during testing.


ayushmanchhabra:
Tests for renaming MacOS Helper apps seem to be flaky. Will address in seperate PR.

PR #1410, opened 2025-07-03
fix(util): update getManifest return type and handle empty response

jssuttles:
Removing the word undefined from the stored data is a weird choice. This changes that.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -10,10 +10,10 @@ import semver from 'semver';
 /**
  * Get manifest (array of NW release metadata) from URL.
  * @param  {string}                      manifestUrl  Url to manifest
- * @returns {Promise<object | undefined>}              - Manifest object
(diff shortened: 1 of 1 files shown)


ayushmanchhabra:
https://github.com/nwutils/nw-builder/blob/548d0e5d09e091ba9eaceaa957dbd49a01becdad/src/util.js#L66
This also needs to be updated, probably why tests are failing

ayushmanchhabra:
Also what is being fixed here? Is this semantic or just a presentational change?

jssuttles:
I would think that running code like the following could lead to bugs.
```javascript
var chunks = undefined;
chunks += '{"data":"stuff"}'; // undefined{"data":"stuff"}
if (chunks !== undefined) chunks = chunks.slice(9); // {"data":"stuff"}
```
So, this is a small refactor.

Issue #1433, opened 2025-08-09
No default value for exec in desktop entry file


T0nd0Tara:
Currently on linux a `.desktop` file should always have the `Exec` key according to the [freedesktop specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html).

As we know the executable name and its path, we could set it as the default if the user didin't enter his custom `exec` and `path` keys

PR #1434, opened 2025-08-09
fix(util): add default value for Exec in Desktop Entry file

T0nd0Tara:
Fixes: #1433


ayushmanchhabra:
These changes should be made here:
https://github.com/nwutils/nw-builder/blob/bf7be13aa55bf192f55272e3801e1cb26171cc1a/src/util.js#L256

T0nd0Tara:
Ok, I've changed it. The reason I didnt do something like

```js
options.app.exec = options.app.exec ?? options.app.name ?? undefined;
if (options.app.exec) {
      options.app.exec = path.resolve(options.app.exec);
}
```

is because I didn't want it to `path.resolve` the `option.app.name` as it wouldn't be a valid path.

I've noticed the `Exec` option of the `.desktop` format doesn't have to be an absolute path so it'll be fine

PR #1577, opened 2026-04-25
fix: do not replace but merge Node manifest options into default options

townofdon:
* Overrides options correctly if the node manifest is present.

Currently, if the node manifest is provided via `package.json`, all of the default options get blown away due to this mutation:

https://github.com/nwutils/nw-builder/blob/becde28fadd6c5dcd8f2d778a5d94ff47a70d9c9/src/index.js#L58

This change applies the manifest options as an override on top of the existing options.


ayushmanchhabra:
~Thanks! Need a test for this, this issue keeps coming back again and again~
I think this was done intentionally, I'm not sure. Will look into the git history.

Update: Got it, this removes all the default options as you mentioned.

PR #1585, opened 2026-04-28
feat(v3): add osxarm64 (Apple Silicon) platform support

sibidharan:
## Summary

The legacy `NwBuilder` class (v3 API) does not support building for macOS Apple Silicon (arm64). NW.js has shipped arm64 macOS builds since ~v0.64, and the newer `get()` API already accepts `arch: "arm64"` — but the v3 `NwBuilder` class never wired this through.

The root cause is that arch parsing is hardcoded as `"x" + platformName.slice(-2)` throughout the codebase, which produces `"x64"` for any platform name ending in `64`. For `osxarm64`, this gives `platform = "osxarm"` and `arch = "x64"` — wrong on both counts.

### Changes

- **`src/constants/Platform.js`** — Add `OSX_ARM64: "osxarm64"` to the Platform enum
- **`src/constants/Platforms.js`** — Add `osxarm64` platform definition with correct `versionNameTemplate` (`osx-arm64.zip`)
- **`src/utilities/detectCurrentPlatform.js`** — Return `OSX_ARM64` when `process.arch === "arm64"` on Darwin
- **`lib/index.cjs`** — Introduce `parsePlatformArch()` helper to replace fragile slice patterns in `downloadNwjs()`, `copyNwjs()`, `init()`, and `runApp()`. Add `osxarm64` to all Mac platform checks (`handleMacApp`, `mergeAppFiles`). Guard against undefined `_version` in `handleMacApp`.
- **`lib/Version.cjs`** — Handle `osxarm64` → `osx-arm64` in `fixPlatformName()`
- **`lib/versions.cjs`** — Map `osx-arm64` from NW.js manifest to `osxarm64` in `mapFilesToPlatforms()`
- **`test/unit/detectCurrentPlatform.js`** — Add unit test for ARM64 platform detection

### Usage

```js
const nw = new NwBuilder({
  files: ['./src/**', './package.json'],
  platforms: ['osxarm64'],
  version: '0.86.0',
  flavor: 'sdk',
});
nw.build();
```

Or via CLI: `--platforms=osxarm64`

### Tested

Built and ran a full NW.js application (Popcorn Time 0.5.1) on an Apple Silicon Mac. Binary confirmed as `Mach-O 64-bit executable arm64`, running natively (not under Rosetta).

## Test plan

- [x] Existing unit tests pass (`node --test test/unit/detectCurrentPlatform.js` — 7/7)
- [x] New ARM64 test case passes
- [x] Full build + run verified on M-series Mac

### DIFF
--- a/lib/Version.cjs
+++ b/lib/Version.cjs
@@ -39,6 +39,7 @@ module.exports = function Version(args) {
     };
   } else {
     var fixPlatformName = function (platform) {
+      if (platform === "osxarm64") return "osx-arm64";
       return platform.replace("32", "-ia32").replace("64", "-x64");
     };
 
--- a/lib/index.cjs
+++ b/lib/index.cjs
@@ -25,6 +25,16 @@ const {
   validate,
 } = require("../dist/index.cjs");
 
+function parsePlatformArch(os) {
+  if (os === "osxarm64") {
+    return { platform: "osx", arch: "arm64" };
+  }
+  return {
+    platform: os.slice(0, os.length - 2),
+    arch: "x" + os.slice(os.length - 2),
+  };
+}
+
 class NwBuilder {
   constructor(options) {
     this.init(options).then(() => {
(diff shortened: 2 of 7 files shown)


ayushmanchhabra:
I appreciate the time and effort you put into this. But v3 has been deprecated for almost 2 years, please use v4 instead. I'll update the npm stable tag to v4 in the next release so there's no confusion.

Feel free to open an issue/PR for v4 (`main` branch) if you run into any issues.

PR #1614, opened 2026-06-29
Fixes glob-enabled relative path bug

### LINKED ISSUE #1612: Glob Folder Structure Bug
When using glob = true: When using a srcDir that is in a different folder than the build folder, files are copied to the wrong location.

```js
glob: true,
srcDir: '../sample/**/**',
outDir: './build',
```

When using the above configuration, the following lines cause a problem:
https://github.com/nwutils/nw-builder/blob/4e43ddce88adbda8a8a2f919aa0410644cece889/src/bld.js#L152-L155

The `../` from the files captured by glob from the srcDir are put into the copy statement. This leads to the files being copied directly to the outDir instead of package.nw.

Furin:
* Fixes an issue with how relative paths are resolved to create the nwProjectDir upon building when using globs. Specifically when traveling up the folder structure.

Fixes: #1612

### DIFF
--- a/src/bld.js
+++ b/src/bld.js
@@ -151,7 +151,7 @@ async function bld({
         file,
         path.resolve(
           nwProjectDir,
-          file,
+          (file.substring(0, 3) == '..' + path.sep ? '.' + path.sep + file.substring(file.indexOf(path.sep,3)) : file),
         ),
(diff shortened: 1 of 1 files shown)


ayushmanchhabra:
@Furin Could you write a test for this?

